Issue #6, opened 2016-05-26
Separate connection errors from errorlog


arkrohne:
The ErrorLog.txt log is often filled with connection issues of failing or intermittent devices. Separating those out into a second connection error log would simplify finding configuration or operational errors.

ritchiecarroll:
How about a flag to optionally log these at all?

arkrohne:
That may be the simplest solution. Sometimes it's useful to see if an error occurred near a connect/disconnect event, but you often don't need that data for the whole time.

rmc00:
@ritchiecarroll Would this be a flag on the adapter level or something more global? Seems like it could be implemented in InputAdapterBase and OutputAdapterBase on the reconnect timer logic.

ritchiecarroll:
@rmc00 that might work, but I think to reduce noise in this case simply implementing some error filter right inside the PhasorMeasurementMapper adapter would suffice. Do you want to take a stab at it?

rmc00:
Yeah, I thought I might give it a shot.

ritchiecarroll:
Feature completed.

PR #8, opened 2016-06-23
Added new property to allow connection errors to not be logged in pha…

rmc00:
…sor measurement mapper.


rmc00:
For issue #6 to enable/disable connection errors in phasor measurement mapper

StephenCWills:
It looks like you forgot to set the variable to its default value of true.

rmc00:
Thanks! It was covered for phasor measurement mapper by the defaultvalue attribute but not for other input adapters. Appreciate your time to look it over.

StephenCWills:
It looks good. Thanks for the contribution!

FYI, the DefaultValue attribute does not actually ensure that the value is initialized with its default. It only affects what the openPDC Manager displays as the default.

PR #10, opened 2016-10-07
Add documentation for Filter Expressions.

EEparker:
I'm not sure if this is the right location for this information, but we have spent some time working with it so I thought other people might find it useful.


ritchiecarroll:
Awesome!

PR #53, opened 2019-11-18
Mono fix

zobo:
Collection of fixes to make project compile with Mono/Linux again.


StephenCWills:
Thanks so much for the pull request! I've confirmed the forward slash works on Windows so no worries there.

zobo:
Great. I will provide similar PR for openPDC, as I'm currently attempting to setup a linux+mysql server and windows management environment (with some trouble).

PR #68, opened 2021-04-27
Add logout endpoint for applications using AuthenticationHandler

StephenCWills:
Adds a `/Security/logout` endpoint to applications using [AuthenticationHandler](https://github.com/GridProtectionAlliance/gsf/blob/master/Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs). Logging out of such applications is as easy as navigating to the endpoint.

This PR also updates `Login.cshtml` to make use of this endpoint automatically, eliminating the need to call [SecurityHub.Logout()](https://github.com/GridProtectionAlliance/gsf/blob/a2f461ae55805b9bde5f50bb0dd75f49b3540574/Source/Libraries/GSF.Web/Security/SecurityHub.cs#L112) in every application. Instead, applications can simply navigate to `/@GSF/Web/Security/Views/Login.cshtml?logout` to get a user-friendly logout page.

### DIFF
--- a/Source/Libraries/GSF.Web/Security/AuthenticationOptions.cs
+++ b/Source/Libraries/GSF.Web/Security/AuthenticationOptions.cs
@@ -66,6 +66,11 @@ namespace GSF.Web.Security
         public const string DefaultLoginPage = Resources.Root + "/Security/Views/Login.cshtml";
 
         /// <summary>
+        /// Default value for <see cref="LogoutPage"/>.
+        /// </summary>
+        public const string DefaultLogoutPage = "/Security/logout";
+
+        /// <summary>
         /// Default value for <see cref="AuthTestPage"/>.
         /// </summary>
         public const string DefaultAuthTestPage = "/AuthTest";
@@ -158,6 +163,11 @@ namespace GSF.Web.Security
         /// Gets or sets the login page used as a redirect location when authentication fails.
         /// </summary>
         public string LoginPage { get; set; } = DefaultLoginPage;
+
+        /// <summary>
+        /// Gets or sets the path for the logout page.
+        /// </summary>
+        public string LogoutPage { get; set; } = DefaultLogoutPage;
 
         /// <summary>
         /// Gets or sets the page name used to test user authorization.
@@ -358,6 +368,11 @@ namespace GSF.Web.Security
         /// Gets the login page used as a redirect location when authentication fails.
         /// </summary>
         public string LoginPage => m_authenticationOptions.LoginPage;
+
+        /// <summary>
+        /// Gets the path for the logout page.
+        /// </summary>
+        public string LogoutPage => m_authenticationOptions.LogoutPage;
 
         /// <summary>
         /// Gets the page name used to test user authorization.
--- a/Source/Libraries/GSF.Web/Security/Scripts/Login.js
+++ b/Source/Libraries/GSF.Web/Security/Scripts/Login.js
@@ -170,17 +170,34 @@ function loginComplete(success, response) {
     }
 }
 
-function logoutComplete(success) {
+function logoutComplete(success, response) {
     $("#workingIcon").hide();
     $("#reloginForm").show();
 
-    if (getBool(getParameterByName("sessionCleared"))) {
+    var sessionClearedParameter = getParameterByName("sessionCleared");
+
+    if (sessionClearedParameter && getBool(sessionClearedParameter)) {
         if (success) {
             $("#response").text("Logout Succeeded");
         }
         else {
             $("#response").text("Partial Logout - Session Only");
             $("#message").text("Client session cache cleared but failed to clear browser cached credentials");
+            $("#responsePanel").show();
+        }
+    }
+    else if (response) {
+        if (success && response === "Logout complete") {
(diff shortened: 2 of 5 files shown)


StephenCWills:
FYI, if you hide whitespace changes when reviewing, the changes are easier to see.

clackner-gpa (approved):

Source/Libraries/GSF.Web/Security/AuthenticationOptions.cs
```diff
@@ -65,6 +65,11 @@ public sealed class AuthenticationOptions : Microsoft.Owin.Security.Authenticati
         /// </summary>
         public const string DefaultLoginPage = Resources.Root + "/Security/Views/Login.cshtml";
 
+        /// <summary>
+        /// Default value for <see cref="LogoutPage"/>.
+        /// </summary>
+        public const string DefaultLogoutPage = "/Security/logout";
```
ritchiecarroll, line 71: This is likely fine or intentional, but why does `DefaultLoginPage` include a `Resources.Root` prefix but `DefaultLogutPage` does not?

Source/Libraries/GSF.Web/Security/Views/Login.cshtml
```diff
@@ -63,6 +64,11 @@
             loginPage = options.LoginPage;
         }
 
+        if (!string.IsNullOrWhiteSpace(options.LogoutPage))
```
ritchiecarroll, line 67: Not a big deal - but other code in this JavaScript file do not wrap to new line before `{`

Source/Libraries/GSF.Web/Security/Views/Login.cshtml
```diff
@@ -265,11 +272,11 @@
...
     <script src="@Resources.Root/Shared/Scripts/jquery.js"></script>
     <script src="@Resources.Root/Shared/Scripts/bootstrap.js"></script>
-    <script src="@Resources.Root/Shared/Scripts/gsf.web.client.js"></script>
+    <script src="@Resources.Root/Shared/Scripts/gsf.web.client.js?ver=1"></script>
     <script src="@Resources.Root/Security/Scripts/core.js"></script>
     <script src="@Resources.Root/Security/Scripts/sha256.js"></script>
     <script src="@Resources.Root/Security/Scripts/enc-base64.js"></script>
     <script src="@Resources.Root/Security/Scripts/ntlm.js?ver=1"></script>
-    <script src="@Resources.Root/Security/Scripts/Login.js?ver=1"></script>
+    <script src="@Resources.Root/Security/Scripts/Login.js?ver=2"></script>
```
ritchiecarroll, line 280: 👍

Source/Libraries/GSF.Web/Security/Scripts/Login.js
```diff
@@ -170,11 +170,13 @@ function loginComplete(success, response) {
     }
 }
 
-function logoutComplete(success) {
+function logoutComplete(success, response) {
     $("#workingIcon").hide();
     $("#reloginForm").show();
 
-    if (getBool(getParameterByName("sessionCleared"))) {
+    var sessionClearedParameter = getParameterByName("sessionCleared");
```
ritchiecarroll, line 177: Should be `const` here for new variable instead of `var`

Source/Libraries/GSF.Web/Shared/Scripts/gsf.web.client.js
```diff
@@ -136,7 +136,7 @@ function clearCachedCredentials(securedUrl, successCallback) {
         if (successCallback)
             xhr.onreadystatechange = function() {
                 if (this.readyState === XMLHttpRequest.DONE)
-                    successCallback(this.status === 401 || this.status === 403);
+                    successCallback(this.status === 200, this.responseText);
```
ritchiecarroll, line 139: Guess this was what was causing the issue. Glad you got this fixed.

I think this is first time I've tried this code review thing - it's pretty decent.

ritchiecarroll (changes requested):

Source/Libraries/GSF.Web/Security/AuthenticationOptions.cs
StephenCWills: The login page is a cshtml page that exists as a `GSF.Web` resource. The logout page is just hardcoded into the `AuthenticationHandler`.

Source/Libraries/GSF.Web/Shared/Scripts/gsf.web.client.js
StephenCWills: Actually, this code was not causing any error. I only changed it because the new logout page returns a 200 for success. The old "clearCredentials" page (which was actually just `AuthTest`) would return a 401 after `securityHub.logout()`, which is what we were expecting here. The error we got was because we didn't call `securityHub.logout()` and therefore got a completely unrelated error because we were still authenticated.

Source/Libraries/GSF.Web/Security/Views/Login.cshtml
StephenCWills: Good catch.

ritchiecarroll (approved):

PR #70, opened 2021-08-18
Openid Connect securityProvider

clackner-gpa:
Added openID Connect Security Provider

### DIFF
--- a/Source/Libraries/GSF.Security/GSF.Security.csproj
+++ b/Source/Libraries/GSF.Security/GSF.Security.csproj
@@ -64,11 +64,19 @@
       <HintPath>..\..\Dependencies\NuGet\ExpressionEvaluator.2.0.4.0\lib\net40\ExpressionEvaluator.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="JWT, Version=8.0.0.0, Culture=neutral, PublicKeyToken=6f98bca0f40f2ecf, processorArchitecture=MSIL">
+      <HintPath>..\..\Dependencies\NuGet\JWT.8.2.3\lib\net46\JWT.dll</HintPath>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\..\Dependencies\NuGet\Newtonsoft.Json.9.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
+    </Reference>
     <Reference Include="System" />
     <Reference Include="System.ComponentModel.DataAnnotations" />
     <Reference Include="System.configuration" />
     <Reference Include="System.Core" />
     <Reference Include="System.DirectoryServices" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Runtime.Caching" />
     <Reference Include="System.Runtime.Serialization" />
     <Reference Include="System.Security" />
     <Reference Include="System.Web" />
@@ -78,6 +86,7 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
+    <Compile Include="OIDCSecurityProvider.cs" />
     <Compile Include="AdoSecurityProvider.cs" />
     <Compile Include="AdoSecurityCache.cs" />
     <Compile Include="Model\AccessLog.cs" />
--- a/Source/Libraries/GSF.Security/ISecurityProvider.cs
+++ b/Source/Libraries/GSF.Security/ISecurityProvider.cs
@@ -195,6 +195,14 @@ namespace GSF.Security
         /// <returns>The user role that the specified user <paramref name="role"/> translates to.</returns>
         string TranslateRole(string role);
 
+        /// <summary>
+        /// Performs a translation of the default login page to a different endpoint.
+        /// </summary>
+        /// <param name="loginUrl"> The URI of the login page specified in the AppSettings </param>
+        /// <param name="encodedPath"> The URI requested by the client </param>
+        /// <param name="referrer"> The Referrer as specified in the request header </param>
+        /// <returns> The URI to be redirected to</returns>
+        string TranslateRedirect(string loginUrl, string encodedPath, string referrer);
         #endregion
     }
 }
--- a/Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
+++ b/Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
@@ -1,0 +1,619 @@
+﻿//******************************************************************************************************
+//  OIDCSecurityProvider.cs - Gbtc
+//
+//  Copyright © 2021, Grid Protection Alliance.  All Rights Reserved.
+//
+//  Licensed to the Grid Protection Alliance (GPA) under one or more contributor license agreements. See
+//  the NOTICE file distributed with this work for additional information regarding copyright ownership.
+//  The GPA licenses this file to you under the MIT License (MIT), the "License"; you may
+//  not use this file except in compliance with the License. You may obtain a copy of the License at:
+//
+//      http://www.opensource.org/licenses/MIT
+//
+//  Unless agreed to in writing, the subject software distributed under the License is distributed on an
+//  "AS-IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. Refer to the
+//  License for the specific language governing permissions and limitations.
+//
+//  Code Modification History:
+//  ----------------------------------------------------------------------------------------------------
+//  08/05/2021 - C. Lackner
+//       Generated original version of source code.
+//
+//******************************************************************************************************
+
+using System;
+using System.Collections.Generic;
+using System.Configuration;
+using System.Data;
+using System.Diagnostics;
+using System.Linq;
+using System.Net.Http;
+using System.Runtime.Caching;
+using System.Security;
+using System.Text.RegularExpressions;
+using System.Threading;
+using GSF.Collections;
+using GSF.Configuration;
+using GSF.Data;
+using GSF.Diagnostics;
+using GSF.Identity;
+using JWT;
+using JWT.Algorithms;
+using JWT.Exceptions;
+using JWT.Serializers;
+using Newtonsoft.Json;
+
+#pragma warning disable S2068
+
+namespace GSF.Security
+{
+    /// <summary>
+    /// Represents an <see cref="ISecurityProvider"/> that uses openID Connect
+    /// </summary>
+    /// <example>
+    /// Required config file entries (automatically added):
+    /// <code>
+    /// <![CDATA[
+    /// <?xml version="1.0"?>
+    /// <configuration>
+    ///   <configSections>
+    ///     <section name="categorizedSettings" type="GSF.Configuration.CategorizedSettingsSection, GSF.Core" />
+    ///   </configSections>
+    ///   <categorizedSettings>
+    ///     <securityProvider>
+    ///       <add name="ProviderType" value="GSF.Security.OIDCSecurityProvider, GSF.Security" description="The type to be used for enforcing security."
+    ///         encrypted="false" />
+    ///       <add name="ClientID" value="xxxx-xxxx-xxxx" description="Defines the ClientID as required per OpenID Connect Standard." encrypted="false" />
+    ///       <add name="Scope" value="user" description="Defines the Scope as required per OpenID Connect Standard." encrypted="false" />
+    ///       <add name="AuthorizationEndpoint" value="user" description="Defines the Endpoint to redirect the user for Authorization." encrypted="false" />
+    ///       <add name="RedirectURI" value="https://localhost:8986/" description="Defines the URI the User get's redirected to after signing in." encrypted="false" />
+    ///       <add name="ClientSecret" value="sssss-ssssss-sssss" description="Defines the Client Secret to encrypt User Information." encrypted="false" />
+    ///       <add name="TokenEndpoint" value="user" description="Defines the Endpoint to get the User Token from." encrypted="false" />
+    ///     </securityProvider>
+    ///   </categorizedSettings>
+    /// </configuration>
+    /// ]]>
+    /// </code>
+    /// </example>
+
+    public class OIDCSecurityProvider : SecurityProviderBase
+    {
+        #region [ Members ]
+
+        // Constants
+        private const string ResponseType = "code";                          // Response Type the SecurityProvider expects
+        //private const string ResponseMode = "form_post";                     // The mode used to get ID Token for the end-user
+        private int NonceSlidingExpiration = 600;                            // Time in Seconds that a Nonce is valid for. This prevents replay attacks
+        // Fields
+        private string m_ClientID;
+        private string m_AuthorizationEndpoint;
+        private string m_TokenEndpoint;
+        private string m_Scope;
+        private string m_RedirectURI;
+        private string m_ClientSecret;
+        private string m_RolesClaim;
+
+        /// <summary>
+        /// Defines the provider ID for the <see cref="AdoSecurityProvider"/>.
+        /// </summary>
+        public new const int ProviderID = 3;
+
+        /// <summary>
+        /// The ClienID used to identify this Application with the Authorization Server
+        /// </summary>
+        public string ClientID
+        {
+            get
+            {
+                return m_ClientID;
+            }
+            set
+            {
+                m_ClientID = value;
+            }
+        }
+
+        /// <summary>
+        /// The Scope used to obtain UserInformation from the Authorization Server
+        /// </summary>
+        public string Scope
+        {
+            get
+            {
+                return m_Scope;
+            }
+            set
+            {
+                m_Scope = value;
+            }
+        }
+
+        /// <summary>
+        /// The Endpoint used to redirect the User
+        /// </summary>
+        public string AuthorizationEndpoint
+        {
+            get
+            {
+                return m_AuthorizationEndpoint;
+            }
+            set
+            {
+                m_AuthorizationEndpoint = value;
+            }
+        }
+
+        /// <summary>
+        /// The Endpoint to get the User Token
+        /// </summary>
+        public string TokenEndpoint
+        {
+            get
+            {
+                return m_TokenEndpoint;
+            }
+            set
+            {
+                m_TokenEndpoint = value;
+            }
+        }
+
+        /// <summary>
+        /// The URI the User get's redirected to after signing in.
+        /// </summary>
+        public string RedirectURI
+        {
+            get
+            {
+                return m_RedirectURI;
+            }
+            set
+            {
+                m_RedirectURI = value;
+            }
+        }
+
+        /// <summary>
+        /// The ClientSecret used to encrypt the user data
+        /// </summary>
+        public string ClientSecret
+        {
+            get
+            {
+                return m_ClientSecret;
+            }
+            set
+            {
+                m_ClientSecret = value;
+            }
+        }
+
+        /// <summary>
+        /// The Claim used to get the Roles for the user
+        /// </summary>
+        public string RolesClaim
+        {
+            get
+            {
+                return m_RolesClaim;
+            }
+            set
+            {
+                m_RolesClaim = value;
+            }
+        }
+
+        private class TokenResponse
+        {
+            public string access_token { get; set; }
+            public string id_token { get; set; }
+        }
+        #endregion
+
+        #region [ Constructor ]
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AdoSecurityProvider"/> class.
+        /// </summary>
+        /// <param name="username">Name that uniquely identifies the user.</param>
+        public OIDCSecurityProvider(string username)
+            : this(username, true, false, false)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AdoSecurityProvider"/> class.
+        /// </summary>
+        /// <param name="username">Name that uniquely identifies the user.</param>
+        /// <param name="canRefreshData">true if the security provider can refresh <see cref="UserData"/> from the backend data store, otherwise false.</param>
+        /// <param name="canResetPassword">true if the security provider can reset user password, otherwise false.</param>
+        /// <param name="canChangePassword">true if the security provider can change user password, otherwise false.</param>
+        protected OIDCSecurityProvider(string username, bool canRefreshData, bool canResetPassword, bool canChangePassword)
+            : base(username, canRefreshData, canResetPassword, canChangePassword)
+        {
+        }
+
+        #endregion
+
+        #region [ Properties ]
+
+        /// <summary>
+        /// Gets last exception reported by the <see cref="AdoSecurityProvider"/>.
+        /// </summary>
+        public Exception LastException
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Gets or sets flag that determines if <see cref="LogAuthenticationAttempt"/> and <see cref="LogError"/> should
+        /// write to the database. Defaults to <c>true</c>.
+        /// </summary>
+        /// <remarks>
+        /// Setting this flag to <c>false</c> may be necessary in cases where a database has been setup to use authentication
+        /// but does not include an "AccessLog" or "ErrorLog" table.
+        /// </remarks>
+        public bool UseDatabaseLogging
+        {
+            get;
+            set;
+        }
+
+        #endregion
+
+        #region [ Methods ]
+
+        /// <summary>
+        /// Loads saved security provider settings from the config file if the <see cref="SecurityProviderBase.PersistSettings"/> property is set to true.
+        /// </summary>
+        /// <exception cref="ConfigurationErrorsException"><see cref="SecurityProviderBase.SettingsCategory"/> has a value of null or empty string.</exception>
+        public override void LoadSettings()
+        {
+            base.LoadSettings();
+
+            // Make sure default settings exist
+            ConfigurationFile config = ConfigurationFile.Current;
+            CategorizedSettingsElementCollection settings = config.Settings[SettingsCategory];
+
+            settings.Add("ClientID", "", "Defines the ClientID as required per OpenID Connect Standard.");
+            settings.Add("Scope", "", "Defines the Scope as required per OpenID Connect Standard.");
+            settings.Add("AuthorizationEndpoint", "", "Defines the Endpoint to redirect the user for Authorization.");
+            settings.Add("RedirectURI", "", "Defines the URI the User get's redirected to after signing in.");
+            settings.Add("ClientSecret", "", "Defines the Client Secret as required per OpenID Connect Standard for decrypting User Information.");
+            settings.Add("TokenEndpoint", "", "Defines the Endpoint to use to grab the User Token.");
+            settings.Add("RolesClaim", "roles", "Defines the claim used to identify the users roles.");
+            
+
+            m_ClientID = settings["ClientID"].ValueAs(m_ClientID);
+            m_Scope = settings["Scope"].ValueAs(m_Scope);
+            m_AuthorizationEndpoint = settings["AuthorizationEndpoint"].ValueAs(m_AuthorizationEndpoint);
+            m_RedirectURI = settings["RedirectURI"].ValueAs(m_RedirectURI);
+            m_ClientSecret = settings["ClientSecret"].ValueAs(m_ClientSecret);
+            m_TokenEndpoint = settings["TokenEndpoint"].ValueAs(m_TokenEndpoint);
+            m_RolesClaim = settings["RolesClaim"].ValueAs(m_RolesClaim);
+        }
+
+        
+        /// <summary>
+        /// Refreshes the <see cref="UserData"/>.
+        /// </summary>
+        /// <returns>true if <see cref="SecurityProviderBase.UserData"/> is refreshed, otherwise false.</returns>
+        public override bool RefreshData()
+        {
+            if (string.IsNullOrEmpty(UserData.Username))
+                return false;
+
+            try
+            {
+                return false;
+            }
+            catch (Exception ex)
+            {
+                LastException = ex;
+                LogError(ex.Source, ex.ToString());
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// Authenticates the user.
+        /// </summary>
+        /// <returns>true if the user is authenticated, otherwise false.</returns>
+        public override bool Authenticate()
+        {
+            IsUserAuthenticated = false;
+            TokenResponse token = null;
+            AuthenticationFailureReason = null;
+            Exception authenticationException = null;
+
+            // Test for pre - authentication failure modes such as invalid Tokens
+            if (String.IsNullOrEmpty(UserData.LoginID))
+            {
+                AuthenticationFailureReason = $"No User authorization code is defined.";
+                IsUserAuthenticated = false;
+            }
+
+            if (!String.IsNullOrEmpty(UserData.LoginID))
+                token = GetToken(UserData.LoginID);
+
+            if (token == null)
+            {
+                AuthenticationFailureReason = $"No Token is available.";
+                IsUserAuthenticated = false;
+            }
+
+            // Update the UserData object with new authentication state
+            if (token != null && DecodeToken(token))
+                IsUserAuthenticated = true;
+
+            try
+            {
+                // Log user authentication result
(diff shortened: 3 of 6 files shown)


Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -149,8 +149,12 @@ protected override Task<AuthenticationTicket> AuthenticateCoreAsync()
             {
                 // Pick the appropriate authentication logic based
                 // on the authorization type in the HTTP headers
+                // or in the URI Parameters if it is using OIDC.
                 if (authorization?.Scheme == "Basic")
                     securityPrincipal = AuthenticateBasic(authorization.Parameter);
+                // If the resources contains a code make an Attempt to Authorize via OIDC Auth server
+                else if (Request.QueryString.HasValue && Request.QueryString.Value.Contains("code="))
```
ritchiecarroll, line 156: This has a high probability of becoming a false positive (same below). Like better to parse QueryString parameters into a HashSet or Dictionary (function likely already exists) and do a `.Contains("code")` instead. Even then, the odds of just having a "code" parameter are high. Is there any other distinguishing characteristic that indicates OIDC, like a header value? Or is the code formatted in such a way that it has a detectible pattern?

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -228,7 +240,9 @@ public override async Task<bool> InvokeAsync()
                 base64Path = Convert.ToBase64String(pathBytes);
                 encodedPath = WebUtility.UrlEncode(base64Path);
 
-                Response.Redirect($"{Options.LoginPage}?redir={encodedPath}{referrer}");
+                ISecurityProvider securityProvider = SecurityProviderCache.CreateProvider("", autoRefresh: false);
+
+                Response.Redirect(securityProvider.TranslateRedirect(Options.LoginPage,encodedPath,referrer));
```
ritchiecarroll, line 245: Nit, add proper spacing:
```c#
Response.Redirect(securityProvider.TranslateRedirect(Options.LoginPage, encodedPath, referrer));
```

Source/Libraries/GSF.Security/packages.config
```diff
@@ -2,4 +2,6 @@
 <packages>
   <package id="Antlr" version="3.5.0.2" targetFramework="net46" />
   <package id="ExpressionEvaluator" version="2.0.4.0" targetFramework="net46" />
+  <package id="JWT" version="8.2.3" targetFramework="net46" />
```
ritchiecarroll, line 5: Now that we have added two new dependencies `GSF.Security` all dependent downstream application installers will need to add new DLLs. `Netwonsoft.Json` should already be in most systems, any that depend on `GSF.Web` anyway, but I think `JWT` is new.

Please verify that the `Netwonsoft.Json` version matches what is deployed with `GSF.Web`.

Also, will need to add `JWT` to the added dependencies in the monthly security report.

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+    public class OIDCSecurityProvider : SecurityProviderBase
+    {
+        #region [ Members ]
+
+        // Constants
+        private const string ResponseType = "code";                          // Response Type the SecurityProvider expects
+        //private const string ResponseMode = "form_post";                     // The mode used to get ID Token for the end-user
+        private int NonceSlidingExpiration = 600;                            // Time in Seconds that a Nonce is valid for. This prevents replay attacks
+        // Fields
+        private string m_ClientID;
```
ritchiecarroll, line 88: Use pascal casing for these member variables starting with first letter beyond `m_` being lower case, e.g., `m_clientID`

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+    /// </example>
+
+    public class OIDCSecurityProvider : SecurityProviderBase
+    {
+        #region [ Members ]
+
+        // Constants
+        private const string ResponseType = "code";                          // Response Type the SecurityProvider expects
+        //private const string ResponseMode = "form_post";                     // The mode used to get ID Token for the end-user
+        private int NonceSlidingExpiration = 600;                            // Time in Seconds that a Nonce is valid for. This prevents replay attacks
```
ritchiecarroll, line 86: Should this be a `const`?

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+
+        /// <summary>
+        /// Defines the provider ID for the <see cref="AdoSecurityProvider"/>.
+        /// </summary>
+        public new const int ProviderID = 3;
+
+        /// <summary>
+        /// The ClienID used to identify this Application with the Authorization Server
+        /// </summary>
+        public string ClientID
```
ritchiecarroll, line 104: These could be auto-properties, e.g.:
```c#
public string ClientID { get; set; }
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+        /// <exception cref="ConfigurationErrorsException"><see cref="SecurityProviderBase.SettingsCategory"/> has a value of null or empty string.</exception>
+        public override void LoadSettings()
+        {
+            base.LoadSettings();
+
+            // Make sure default settings exist
+            ConfigurationFile config = ConfigurationFile.Current;
+            CategorizedSettingsElementCollection settings = config.Settings[SettingsCategory];
+
+            settings.Add("ClientID", "", "Defines the ClientID as required per OpenID Connect Standard.");
```
ritchiecarroll, line 279: Use `nameof` for named properties:
```c#
settings.Add(nameof(ClientID), "", "Defines the ClientID as required per OpenID Connect Standard.");
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+
+        
+
+        /// <summary>
+        /// Logs user authentication attempt.
+        /// </summary>
+        /// <param name="loginSuccess">true if user authentication was successful, otherwise false.</param>
+        protected virtual void LogAuthenticationAttempt(bool loginSuccess)
+        {
+
```
ritchiecarroll, line 382: Nit: spurious new line

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+        /// <summary>
+        /// Logs information about an encountered exception to the backend data store.
+        /// </summary>
+        /// <param name="source">Source of the exception.</param>
+        /// <param name="message">Detailed description of the exception.</param>
+        /// <returns>true if logging was successful, otherwise false.</returns>
+        protected virtual bool LogError(string source, string message)
+        {
+            if (!string.IsNullOrWhiteSpace(SettingsCategory) && !string.IsNullOrWhiteSpace(source) && !string.IsNullOrWhiteSpace(message))
```
ritchiecarroll, line 426: Consider reducing `if` nesting complexity, e.g.:
```c#
if (string.IsNullOrWhiteSpace(SettingsCategory) || string.IsNullOrWhiteSpace(source) || string.IsNullOrWhiteSpace(message))
    return false;
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+        }
+
+       
+        /// <summary>
+        /// Exchange Authorization Code for a Token
+        /// </summary>
+        /// <param name="code"> The Authorization Code returned by the Auth Server</param>
+        private TokenResponse GetToken(string code)
+        {
+
```
ritchiecarroll, line 463: Nit: spurious new line

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+            };
+
+            void ConfigureRequest(HttpRequestMessage request)
+            {
+                request.RequestUri = new Uri($"{m_TokenEndpoint}");
+                request.Method = HttpMethod.Post;               
+                request.Content = new FormUrlEncodedContent(postParams);
+            }
+
+            TokenResponse r;
```
ritchiecarroll, line 479: Move variable declaration closer to use if possible and combine declaration and assignment, e.g.:
```c#
TokenResponse r = JsonConvert.DeserializeObject<TokenResponse>(response.Content.ReadAsStringAsync().Result);
```

Also consider making the variable name more descriptive, e.g., renaming `r` to `token`.

Generally single variable letter usage is reserved for `for` loops according to GPA style guide:
https://gridprotectionalliance.org/docs/GPA_Coding_Guidelines_2011_03.pdf

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+                {
+                    response = client.SendAsync(request).Result;
+                    if (!response.IsSuccessStatusCode)
+                        return null;
+                    r = JsonConvert.DeserializeObject<TokenResponse>(response.Content.ReadAsStringAsync().Result);
+                }
+            }
+
+            return r;
+
```
ritchiecarroll, line 494: Nit: spurious new line - there are many of these - I will stop commenting on these

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+            
+        }
+
+        private bool DecodeToken(TokenResponse token)
+        {
+
+            try
+            {
+                IJsonSerializer serializer = new JsonNetSerializer();
+                var provider = new UtcDateTimeProvider();
```
ritchiecarroll, line 503: Use of `var` is discouraged unless it required by use of anonymous type.

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+        /// <param name="referrer"> The Referrer as specified in the request header </param>
+        /// <returns> The URI to be redirected to</returns>
+        public override string TranslateRedirect(string loginUrl, string encodedPath, string referrer)
+        {
+            byte[] nonce = new byte[16];
+            new Random().NextBytes(nonce);
+
+            s_nonceCache.Add(BitConverter.ToString(nonce).Replace("-", ""), 0, new CacheItemPolicy { SlidingExpiration = TimeSpan.FromSeconds(NonceSlidingExpiration) });
+
+            string redirect = $"{m_AuthorizationEndpoint}?";
```
ritchiecarroll, line 599: Suggest using a `StringBuilder` to concat strings for better performance.

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -350,6 +365,24 @@ private SecurityPrincipal AuthenticatePassthrough()
             return new SecurityPrincipal(securityIdentity);
         }
 
+        // Applies authentication for requests using OpenID Connect authentication.
+        private SecurityPrincipal AuthenticateCode()
+        {
+
```
ritchiecarroll, line 371: Nit: spurious new line

Source/Libraries/GSF.Security/GSF.Security.csproj
```diff
@@ -64,11 +64,19 @@
       <HintPath>..\..\Dependencies\NuGet\ExpressionEvaluator.2.0.4.0\lib\net40\ExpressionEvaluator.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="JWT, Version=8.0.0.0, Culture=neutral, PublicKeyToken=6f98bca0f40f2ecf, processorArchitecture=MSIL">
+      <HintPath>..\..\Dependencies\NuGet\JWT.8.2.3\lib\net46\JWT.dll</HintPath>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <HintPath>..\..\Dependencies\NuGet\Newtonsoft.Json.9.0.1\lib\net45\Newtonsoft.Json.dll</HintPath>
```
ritchiecarroll, line 71: This version of `Newtonsoft` seems wrong - we are using version `12.0.3` elsewhere. We should only use the same version of this DLL since we can only deploy and use one within products that use `GSF`.

Source/Libraries/GSF.Security/ISecurityProvider.cs
```diff
@@ -195,6 +195,14 @@ LogEventFunctionSignature LogEvent
...
 
+        /// <summary>
+        /// Performs a translation of the default login page to a different endpoint.
+        /// </summary>
+        /// <param name="loginUrl"> The URI of the login page specified in the AppSettings </param>
+        /// <param name="encodedPath"> The URI requested by the client </param>
+        /// <param name="referrer"> The Referrer as specified in the request header </param>
+        /// <returns> The URI to be redirected to</returns>
+        string TranslateRedirect(string loginUrl, string encodedPath, string referrer);
         #endregion
```
ritchiecarroll, line 206: Nit: add a new line between code declaration and `#endregion`

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -149,8 +149,12 @@ protected override Task<AuthenticationTicket> AuthenticateCoreAsync()
             {
                 // Pick the appropriate authentication logic based
                 // on the authorization type in the HTTP headers
+                // or in the URI Parameters if it is using OIDC.
                 if (authorization?.Scheme == "Basic")
                     securityPrincipal = AuthenticateBasic(authorization.Parameter);
+                // If the resources contains a code make an Attempt to Authorize via OIDC Auth server
```
ritchiecarroll, line 155: Nit: add a new line for spacing

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -184,6 +188,7 @@ public override async Task<bool> InvokeAsync()
             //  (1) Access resource marked as anonymous - let pipeline continue
             //  (2) Access resource as authenticated user - let pipeline continue
             //  --- remaining use cases are unauthorized ---
+            //  (2a) Access resource as result of a redirect from the OIDC Auth Server
```
ritchiecarroll, line 191: Just renumber these comments instead of adding 2a

ritchiecarroll (changes requested):
Some updates have been suggested.

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
clackner-gpa: I will change the logic to use a Dictionary or hashset. 
The reason I am not too concerned about false positives here is :
if the requested resource is available anonymous it should never get to this point. 
If the user is logged in trying to access a resource it should not get to this point either. 

The general flow for OIDC is User requests resource -> Application responds with Redirect to auth server and generates a key -> Auth Server lets user log in and redirects back to http GET endpoint with `code` parameter (which is this point) 
We could change the logic by requiring that endpoint to be set in the config file instead of looking for the `code` parameter.

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+                try
+                {
+                    if (tokenContent.ContainsKey(m_RolesClaim))
+                        userData.Roles = (List<string>)(tokenContent.GetOrDefault(m_RolesClaim));
+                }
+                catch (Exception ex)
+                {
+                    LastException = ex;
+                    Log.Publish(MessageLevel.Warning, MessageFlags.SecurityMessage, "DecodingTokenError", "Failed to decode Roles Claim.", "Failed to decode Claim for roles.", ex);
+                }
```
StephenCWills, line 542: `tokenContent.GetOrDefault("...").ToString()` looks like a recipe for a `NullReferenceException`.

If the field is expected to be there, consider using the indexer `tokenContent["..."].ToString()` instead. Even better would be to use `TryGetValue()` and throw an appropriate exception.

```c#
if (!tokenContent.TryGetValue("sub", out object sub))
    throw new Exception("Token data was not derived from valid JWT: missing 'sub'");

UserData userData = new UserData(sub.ToString());
```

For optional parameters, you can just flip the condition on the if statement.

```c#
if (tokenContent.TryGetValue("given_name", out object firstName))
    userData.FirstName = firstName.ToString();
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+                { "grant_type", "authorization_code" },
+                { "client_id", m_ClientID },
+                { "client_secret", m_ClientSecret },
+                { "code", code },
+                { "redirect_uri", m_RedirectURI }
+            };
+
+            void ConfigureRequest(HttpRequestMessage request)
+            {
+                request.RequestUri = new Uri($"{m_TokenEndpoint}");
```
StephenCWills, line 474: `m_TokenEndpoint` is already a string so the interpolation is doing nothing.

```suggestion
                request.RequestUri = new Uri(m_TokenEndpoint);
```

StephenCWills (changes requested):

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,619 @@
...
+                // database exceptions can be tracked separately (via LastException property) from other login exceptions, e.g., when
+                // a read-only database is being used or current user only has read-only access to database.
+                if (!string.IsNullOrWhiteSpace(SettingsCategory) && UseDatabaseLogging)
+                {
+                    AdoDataConnection database = new AdoDataConnection(SettingsCategory);
+
+                    using (IDbConnection connection = database.Connection)
+                    {
+                        connection.ExecuteNonQuery(database.ParameterizedQueryString("INSERT INTO AccessLog (UserName, AccessGranted) VALUES ({0}, {1})", "userName", "accessGranted"), UserData.Username, loginSuccess ? 1 : 0);
+                    }
```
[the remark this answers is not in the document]
clackner-gpa: Does this also need to be changed in the `AdoSecurityProvider`? I left these function as they were over there.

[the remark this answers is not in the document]
StephenCWills: I see, so this came from `AdoSecurityProvider`. I suppose that code was probably written before we created the `AdoDataConnection.ExecuteNonQuery()` function.

Technically, there are no errors in the original code. The main differences are:

* In the old code, database parameters are named `@userName` and `@accessGranted` in the underlying SQL statement. In the updated code, they are named `@p0` and `@p1`.
* In the old code, `AdoDataConnection.Dispose()` never gets called. This likely generates a warning from code analysis, but the underlying connection is probably the only disposable resource used by that class.
* The old code explicitly converts the `loginSuccess` Boolean to an integer. The updated code relies on the underlying logic in `AdoDataConnection` which will only convert to an integer if the underlying database engine is Oracle or PostgreSQL.
* The updated code looks better, and provides a better example to follow.

So there's no particular need to update the code in `AdoSecurityProvider`, but it would be nice to clean things up.

Source/Libraries/GSF.Security/packages.config
clackner-gpa: Removed JWT Dependency to simplify Dependency Handling.

clackner-gpa: Updated Newtonsoft to match `GSF.Web`

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
clackner-gpa: Outdated since this function was removed

Source/Libraries/GSF.Security/AdoSecurityProvider.cs
```diff
@@ -780,11 +780,9 @@ protected virtual void LogAuthenticationAttempt(bool loginSuccess)
                 // a read-only database is being used or current user only has read-only access to database.
                 if (!string.IsNullOrWhiteSpace(SettingsCategory) && UseDatabaseLogging)
                 {
-                    AdoDataConnection database = new AdoDataConnection(SettingsCategory);
-
-                    using (IDbConnection connection = database.Connection)
+                    using (AdoDataConnection connection = new AdoDataConnection(SettingsCategory))
                     {
-                        connection.ExecuteNonQuery(database.ParameterizedQueryString("INSERT INTO AccessLog (UserName, AccessGranted) VALUES ({0}, {1})", "userName", "accessGranted"), UserData.Username, loginSuccess ? 1 : 0);
+                        connection.ExecuteNonQuery("INSERT INTO AccessLog (UserName, AccessGranted) VALUES ({0}, {1})", UserData.Username, loginSuccess ? 1 : 0);
```
StephenCWills, line 785: `AdoDataConnection` resolves boolean parameters under the covers.

https://github.com/GridProtectionAlliance/gsf/blob/4d453a383e4387e29ab2dfaeedd9da0bda42d1a6/Source/Libraries/GSF.Core/Data/AdoDataConnection.cs#L885-L896

```suggestion
                        connection.ExecuteNonQuery("INSERT INTO AccessLog (UserName, AccessGranted) VALUES ({0}, {1})", UserData.Username, loginSuccess);
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+        #endregion
+
+        #region [ Constructors ]
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AdoSecurityProvider"/> class.
+        /// </summary>
+        /// <param name="username">Name that uniquely identifies the user.</param>
+        public OIDCSecurityProvider(string username)
+            : this(username, true, false, false)
```
StephenCWills, line 106: This provider's `RefreshData()` method simply returns false so I don't think `CanRefreshData` should be true.

```suggestion
            : this(username, false, false, false)
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+            {
+                return false;
+            }
+            catch (Exception ex)
+            {
+                LastException = ex;
+                LogError(ex.Source, ex.ToString());
+                throw;
+            }
+        }
```
StephenCWills, line 264: This is just returning false in all cases, and no exception could possibly occur.

```suggestion
        public override bool RefreshData()
        {
            return false;
        }
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+        /// <returns>true if the user is authenticated, otherwise false.</returns>
+        public override bool Authenticate()
+        {
+            IsUserAuthenticated = false;
+            TokenResponse token = null;
+            AuthenticationFailureReason = null;
+            Exception authenticationException = null;
+
+            // Test for pre - authentication failure modes such as invalid Tokens
+            if (String.IsNullOrEmpty(UserData.LoginID))
```
StephenCWills, line 278:
```suggestion
            if (string.IsNullOrEmpty(UserData.LoginID))
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+            {
+                AuthenticationFailureReason = $"No User authorization code is defined.";
+                IsUserAuthenticated = false;
+            }
+
+            if (!String.IsNullOrEmpty(UserData.LoginID))
```
StephenCWills, line 284:
```suggestion
            if (!string.IsNullOrEmpty(UserData.LoginID))
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+        /// <summary>
+        /// Authenticates the user.
+        /// </summary>
+        /// <returns>true if the user is authenticated, otherwise false.</returns>
+        public override bool Authenticate()
+        {
+            IsUserAuthenticated = false;
+            TokenResponse token = null;
+            AuthenticationFailureReason = null;
+            Exception authenticationException = null;
```
StephenCWills, line 275:
```suggestion
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+            }
+
+            // If an exception occurred during authentication, rethrow it after logging authentication attempt
+            if ((object)authenticationException != null)
+            {
+                LastException = authenticationException;
+                LogError(authenticationException.Source, authenticationException.ToString());
+                string message = $"Exception occurred during authentication attempt: {authenticationException.Message}";
+                throw new Exception(message, authenticationException);
+            }
```
StephenCWills, line 317: This logic appears to be unnecessary, unless you intended to handle exceptions thrown by `GetToken()` and `DecodeToken()` (which might be a good idea).

```suggestion
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+                {
+                    LogError(ex.Source, ex.ToString());
+                }
+
+                // Attempt to write success or failure to the database - we allow caller to catch any possible exceptions here so that
+                // database exceptions can be tracked separately (via LastException property) from other login exceptions, e.g., when
+                // a read-only database is being used or current user only has read-only access to database.
+                if (!string.IsNullOrWhiteSpace(SettingsCategory) && UseDatabaseLogging)
+                {
+
```
StephenCWills, line 353:
```suggestion
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+            using (HttpRequestMessage request = new HttpRequestMessage())
+            {
+                ConfigureRequest(request);
+               
+                response = Client.SendAsync(request).Result;
+                if (!response.IsSuccessStatusCode)
+                    return null;
+                return JsonConvert.DeserializeObject<TokenResponse>(response.Content.ReadAsStringAsync().Result);
+                
+            }
```
StephenCWills, line 429: This should dispose of the `HttpResponseMessage`. Also, I've read that `task.GetAwaiter().GetResult()` is nearly always preferred because it handles exceptions more intuitively.

```suggestion
            using (HttpRequestMessage request = new HttpRequestMessage())
            {
                ConfigureRequest(request);
               
                using (HttpResponseMessage response = Client.SendAsync(request).GetAwaiter().GetResult())
                {
                    if (!response.IsSuccessStatusCode)
                        return null;
                    return JsonConvert.DeserializeObject<TokenResponse>(response.Content.ReadAsStringAsync().GetAwaiter().GetResult());
                }
            }
```

StephenCWills: Furthermore, it might be a bit more readable, more performant, and more adaptable to future API adjustments if you make this method `async`, use `await` on the tasks, and defer the `task.GetAwaiter().GetResult()` to the caller.

```c#
public override bool Authenticate()
{
    ...
    if (!string.IsNullOrEmpty(UserData.LoginID))
        token = GetTokenAsync(UserData.LoginID).GetAwaiter().GetResult();
    ...
}

private async Task<TokenResponse> GetTokenAsync(string code)
{
    ...
    using (HttpRequestMessage request = new HttpRequestMessage())
    {
        ConfigureRequest(request);
       
        using (HttpResponseMessage response = await Client.SendAsync(request))
        {
            if (!response.IsSuccessStatusCode)
                return null;
            return JsonConvert.DeserializeObject<TokenResponse>(await response.Content.ReadAsStringAsync());
        }
    }
}
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+            {
+                string message = $"Exception occurred while decoding the OpenID token: {ex.Message}";
+                throw new Exception(message, ex);
+            }
+        }
+
+        private JObject DecodeJWT(string token)
+        {
+            if (!token.Contains("."))
+                throw new InvalidExpressionException("A valid JQT token requires at least one '.'");
```
StephenCWills, line 486:
```suggestion
                throw new InvalidExpressionException("A valid JWT token requires at least one '.'");
```

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+
+            string jsoeHeader = token.Substring(0, token.IndexOf("."));
```
StephenCWills, line 488: Is it supposed to be `jsonHeader`?

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
```diff
@@ -0,0 +1,582 @@
...
+                throw new FormatException("JWE Tokens are not supported");
+
+            token = token.Substring(token.IndexOf(".") + 1);
+            string payloadString = token.Substring(0, token.IndexOf("."));
+
+            string signatureString = "";
+            if (token.Contains("."))
+            {
+                token = token.Substring(token.IndexOf(".") + 1);
+                signatureString = token.Substring(0, token.IndexOf("."));
```
StephenCWills, line 509: We keep repeating this substring operation. Maybe it would be better to use `token.Split('.')`? I feel like either the signature or the payload is probably not being parsed correctly because at some point `token.IndexOf(".")` should return `-1`.

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -203,11 +211,21 @@ public override async Task<bool> InvokeAsync()
                 return true; // Abort pipeline
             }
 
+
```
StephenCWills, line 214:
```suggestion
```

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -203,11 +211,21 @@ public override async Task<bool> InvokeAsync()
...
+            // If the resources contains a code make an Attempt to Authorize via OIDC Auth server
+            NameValueCollection queryParameters = System.Web.HttpUtility.ParseQueryString(Request.QueryString.Value);
+
+            if (Request.QueryString.HasValue && queryParameters.AllKeys.Contains("code"))
+                return false;
+
+            
+
+
             // Abort pipeline with appropriate response
```
StephenCWills, line 229:
```suggestion
            // If the resources contains a code make an Attempt to Authorize via OIDC Auth server
            NameValueCollection queryParameters = System.Web.HttpUtility.ParseQueryString(Request.QueryString.Value);

            if (Request.QueryString.HasValue && queryParameters.AllKeys.Contains("code"))
                return false;

            // Abort pipeline with appropriate response
```

Source/Libraries/GSF.Web/Security/AuthenticationHandler.cs
```diff
@@ -335,6 +355,7 @@ private SecurityPrincipal AuthenticatePassthrough()
         {
             string username = Request.User?.Identity.Name;
 
+
```
StephenCWills, line 358:
```suggestion
```

StephenCWills (changes requested):

Source/Libraries/GSF.Security/OIDCSecurityProvider.cs
clackner-gpa: No it's JSOE. 

> 
   JOSE Header
      JSON object containing the parameters describing the cryptographic operations and parameters employed.  The JOSE (JSON Object Signing and Encryption) Header is comprised of a set of Header Parameters.

it is in Json format but this specific header is called JSOE.

StephenCWills: Gotcha. I'd never seen that acronym before.

StephenCWills (approved):

StephenCWills:
@clackner-gpa Alright, I think it's in good shape. You should probably review the `DecodeJWT()` method and also maybe the `Authenticate()` method.

PR #93, opened 2022-01-06
Disable form autocompletion by default on paged view model

StephenCWills:
This solves issues where form autocompletion would autofill credentials for things like field devices and file shares using the browser's credentials. This change disables autocompletion by default so paged view models will have to opt-in to get form autocompletion.

https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion

### DIFF
--- a/Source/Libraries/GSF.Web/Model/Views/PagedViewModel.cshtml
+++ b/Source/Libraries/GSF.Web/Model/Views/PagedViewModel.cshtml
@@ -255,7 +255,7 @@
                 </h4>
             </div>
             <div class="modal-body auto-height" data-bind="with: currentRecord, validationOptions: {messageTemplate: 'validationMessageTemplate'}, css: {'modal-readonly': recordMode()===RecordMode.View}">
-                <form role="form">
+                <form role="form" @((ViewBag.AddNewEditAutocomplete ?? false) ? "" : "autocomplete=\"off\"")>
                     @Raw(ViewBag.AddNewEditDialog)
                 </form>
             </div>


ritchiecarroll (commented):
Nice! However, should this default to a breaking behavior though? Not sure, but the auto-fill might be considered a "nice-feature" on some screens. Then could "opt-out" where deemed appropriate.

Also, perhaps the `AddField` type methods could also apply this at field granularity as well - where removing from all fields full form might be less desirable. Haven't looked, but this might be possible to apply with some of the existing parameter values - otherwise would need to add one.

ritchiecarroll:
I checked - looks like there is not a way to do this currently. Maybe it would make sense to add a parameter like `, string customAttributes` where you could pass in `, customAttributes: "autocomplete='off'"` when adding a field?

StephenCWills:
I set it up this way because I can't imagine a scenario in which we would be asking for a user's autocomplete info on one of these add-new/edit forms. Even if the user has a bunch of devices with the same credentials, for instance, it's reasonable to assume they will have different types of devices with different credentials so it would be difficult for the browser to fill in the appropriate value. I think `autocomplete="off"` is simply the most sane default. Also potentially worth noting...

> Setting `autocomplete="off"` on fields has two effects:
>
> * It tells the browser not to save data inputted by the user for later autocompletion on similar forms, though heuristics for complying vary by browser.
> * It stops the browser from caching form data in the session history. When form data is cached in session history, the information filled in by the user is shown in the case where the user has submitted the form and clicked the Back button to go back to the original form page.

> If a site sets `autocomplete="off"` for a `<form>`, and the form includes username and password input fields, then the browser still offers to remember this login, and if the user agrees, the browser will autofill those fields the next time the user visits the page.

So apparently the browser still offers to remember for specific fields, but theoretically will no longer apply heuristics to fill in data on "similar" forms.

ritchiecarroll:
Yeah - it's just this paged view model is used everywhere, was more worried about one of our end users being "accustomed" to having that as an option on one of the screens. But since there's an option to "bring it back", guess what you have will fix more issues than it may possibly cause.

StephenCWills:
FYI, I just tested this in Edge, and it's not working properly so we should hold off on merging.

ritchiecarroll (approved):
I think this will work fine. FYI, still think would like to add `customAttributes` option to `AddField` functions at some point for more finite control.

ritchiecarroll:
> FYI, I just tested this in Edge, and it's not working properly so we should hold off on merging.

Yikes. OK

StephenCWills:
Edge apparently doesn't honor the `autocomplete="off"` attribute, and their heuristic is still blindly autofilling whatever it feels like so I'll just close this. Back to the drawing board.

PR #104, opened 2022-11-13
Cleaned up auth checks and try-catch

clackner-gpa:
Simplified Authentication checks and removed unnecessary try-catch


Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -162,29 +162,22 @@ public ModelController()
         [HttpGet, Route("New")]
         public virtual IHttpActionResult GetNew()
```
StephenCWills, line 163: It looks to me like you removed the try-catch from every function except this one.

StephenCWills (approved):

PR #111, opened 2023-01-04
Fix for the auth check on searchable lists, was returning unauthorize…

gcsantos-gpa:
…d when authorized.

Compare to other http methods.

### DIFF
--- a/Source/Libraries/GSF.Web/Model/ModelController.cs
+++ b/Source/Libraries/GSF.Web/Model/ModelController.cs
@@ -395,7 +395,7 @@ namespace GSF.Web.Model
         [HttpPost, Route("{parentID?}/SearchableList")]
         public virtual IHttpActionResult GetSearchableList([FromBody] PostData postData, string parentID = null)
         {
-            if (!GetAuthCheck() && !AllowSearch)
+            if (!GetAuthCheck() || !AllowSearch)
                 return Unauthorized();
 
             if (ParentKey != string.Empty && parentID != null)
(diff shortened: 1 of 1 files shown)


Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -378,7 +378,7 @@ public virtual IHttpActionResult Delete(T record)
         [HttpPost, Route("SearchableList")]
         public virtual IHttpActionResult GetSearchableList([FromBody] PostData postData)
         {
-            if (GetAuthCheck() && !AllowSearch)
+            if (!GetAuthCheck() && !AllowSearch)
```
clackner-gpa, line 381: That Logic will require a user to be authorized for `GET` OR `AllowSearch` to be true for it to continue since either `GetAuthCheck()` or `AllowSerach` being `true` will cause `Unauthorized`

The idea here was the user needs to be authorized for GET AND the model has to bel searchable. 

```suggestion
            if (!GetAuthCheck() || !AllowSearch)
```

Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -395,7 +395,7 @@ public virtual IHttpActionResult GetSearchableList([FromBody] PostData postData)
         [HttpPost, Route("{parentID?}/SearchableList")]
         public virtual IHttpActionResult GetSearchableList([FromBody] PostData postData, string parentID = null)
         {
-            if (GetAuthCheck() && !AllowSearch)
+            if (!GetAuthCheck() && !AllowSearch)
```
clackner-gpa, line 398: see above
```suggestion
            if (!GetAuthCheck() || !AllowSearch)
```

clackner-gpa (changes requested):

Source/Libraries/GSF.Web/Model/ModelController.cs
clackner-gpa: Note that the issue you encountered should probably be resolved by adding `AllowSearch` to the model since that would have made the old logic work properly. There is still an issue with the old logic approving users without `GET` access tho

PR #117, opened 2023-02-23
Allow override for Query functions

clackner-gpa:
This is needed to add additional Logic in Sec urityGroup Controller to pull Display Name from AD/Azure


ritchiecarroll (approved):
Looks fine

PR #121, opened 2023-03-18
Added Filter to remove Application roles not associated wioth NodeID

clackner-gpa:


Source/Libraries/GSF.Security/AdoSecurityProvider.cs
```diff
@@ -574,7 +574,30 @@ public override bool RefreshData()
 
                 // Filter explicitly assigned application roles for current user - this will return an empty set if no
                 // explicitly defined roles exist for the user -or- user doesn't exist in the database.
-                DataRow[] userApplicationRoles = securityContext.Tables[ApplicationRoleUserAccountTable].Select($"UserAccountID = '{EncodeEscapeSequences(userAccountID.ToString())}'");
+                DataRow[] userApplicationRoles = securityContext.Tables[ApplicationRoleUserAccountTable]
+                    .Select($"UserAccountID = '{EncodeEscapeSequences(userAccountID.ToString())}'");
+
+                //Filter ApplicationRoles by NodeID in case multiple Nodes use the same database
+                Func<DataRow, bool> nodeIDFilter = (row) =>
```
ritchiecarroll, line 581: Traditionally, user application roles have crossed NodeID boundaries. This changes that notion. That may be OK since we do not normally use NodeIDs for the way they were originally intended. However, I wonder if this should be a configurable option, defined in the security config file settings, that can default to true for your desired use case, but allow an implementation to opt-out of the behavior in case there are deployed systems that may expect the original behavior?

Source/Libraries/GSF.Security/AdoSecurityProvider.cs
```diff
@@ -574,7 +574,30 @@ public override bool RefreshData()
 
                 // Filter explicitly assigned application roles for current user - this will return an empty set if no
                 // explicitly defined roles exist for the user -or- user doesn't exist in the database.
-                DataRow[] userApplicationRoles = securityContext.Tables[ApplicationRoleUserAccountTable].Select($"UserAccountID = '{EncodeEscapeSequences(userAccountID.ToString())}'");
+                DataRow[] userApplicationRoles = securityContext.Tables[ApplicationRoleUserAccountTable]
+                    .Select($"UserAccountID = '{EncodeEscapeSequences(userAccountID.ToString())}'");
+
+                //Filter ApplicationRoles by NodeID in case multiple Nodes use the same database
```
ritchiecarroll, line 580: Nit-picky thing: Add space after `//` so: `// Filter

Source/Libraries/GSF.Security/AdoSecurityProvider.cs
```diff
@@ -574,7 +574,30 @@ public override bool RefreshData()
...
+                Func<DataRow, bool> nodeIDFilter = (row) =>
+                {
+                    DataRow applicationRole = null;
+
+                    if (securityContext.Tables[ApplicationRoleTable].PrimaryKey.Length > 0)
+                        applicationRole = securityContext.Tables[ApplicationRoleTable].Rows.Find(row["ApplicationRoleID"]);
```
ritchiecarroll, line 586: Nit-picky thing: Add braces around this initial `if` result, our coding style spec says that if you use braces around one section, you should use it around the other.

Source/Libraries/GSF.Security/AdoSecurityProvider.cs
```diff
@@ -574,7 +574,30 @@ public override bool RefreshData()
...
+                    else
+                    {
+                        DataRow[] applicationRoles = securityContext.Tables[ApplicationRoleTable].Select($"ID = '{EncodeEscapeSequences(row["ApplicationRoleID"].ToString())}'");
+
+                        if (applicationRoles.Length > 0)
+                            applicationRole = applicationRoles[0];
+                    }
+
+                    if (applicationRole is null || Convert.IsDBNull(applicationRole["Name"]))
```
ritchiecarroll, line 595: This can be a single return statement, like:
```c#
return applicationRole is not null && !Convert.IsDBNull(applicationRole["Name"]);
```

Source/Libraries/GSF.Security/AdoSecurityProvider.cs
StephenCWills: Is this not simply a bug? `ApplicationRole` has a `NodeID` so roles from another node assigned to a user shouldn't count when evaluating role membership on this node. If we don't filter them here, then we'll skip the lookup for implicit role assignment and then find out later that the user doesn't actually have an explicit role assigned on this node.

clackner-gpa: > If we don't filter them here, then we'll skip the lookup for implicit role assignment and then find out later that the user doesn't actually have an explicit role assigned on this node.

That is exactly what happened. IF user application roles cross NodeID boundaries then why do we associated the NodeID in the `ApplicationRole` Table? the fact that we do include `NodeID` implies to me that it only applies to a  single Node. There are also other places in the code that make that assumptions causing the issue Stephen mentioned since AUthentication now says you have roles but the rest of the code won't match that.

ritchiecarroll: Yeah, this notion goes "way back" to when these tables were created by Pinal at TVA. I suppose the risk is limited and perhaps treat it like a "bug" as Stephen mentions.

PR #126, opened 2023-04-27
Load implementations updates

ritchiecarroll:


Source/Libraries/GSF.TimeSeries/Reports/ReportingProcessCollection.cs
```diff
@@ -89,6 +90,56 @@ public virtual string Status
...
+        {
+            // Manually load known reporting processes as an optimization
+            Add(new CompletenessReportingProcess());
+            Add(new CorrectnessReportingProcess());
+
+            foreach (IReportingProcess reportingProcess in this)
+                newReportProcessHandler(reportingProcess);
+
+            // Load any user defined reporting processes on a background thread
+            new Thread(() =>
```
StephenCWills, line 108: Exceptions thrown from `newReportProcessHandler` or `exceptionHandler` could end up crashing the whole process. Might be good to have a catch-all in the lambda to log exceptions instead.

StephenCWills (approved):

PR #132, opened 2023-06-10
Added Pageination

clackner-gpa:
This Adds 2 pagination Endpoints to the Model Controller. 
I did not bother adding one if no Filters are present (equivalent of `Get()`) since we can always supply an empty `Searches` It does mean there is only `POST` support when doing Pageination but I figure that would be fine.

### DIFF
--- a/Source/Libraries/GSF.Web/Model/ModelController.cs
+++ b/Source/Libraries/GSF.Web/Model/ModelController.cs
@@ -65,6 +65,14 @@ namespace GSF.Web.Model
             public string OrderBy { get; set; }
             public bool Ascending { get; set; }
         }
+
+        public class PagedResults
+        {
+            public int RecordPerPage { get; set; }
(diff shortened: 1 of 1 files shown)


StephenCWills (approved):
The main thing I fixed were the calls to `Math.Ceil()`, which doesn't exist. The function is called `Math.Ceiling()`, and it returns `double`. So it should have been `(int)Math.Ceiling(recordCount / (double)(Take ?? 50))`, but I thought maybe it would be clearer to eliminate the type conversions by doing the ceiling using integer arithmetic. If you think it's unclear, feel free to change it back.

I'm wondering how the build check was able to pass when the build should have been failing due to `Math.Ceil()`. May need to look into that.

EDIT: @prestoncraw looked into the build check and figured out what the issue was. It had been building the master branch instead of the PR branch.

PR #134, opened 2023-06-30
Skip NULL values in JoinKeyValuePairs

clackner-gpa:
After discussing with Stephen this seemed like the best course of action if a value is `NULL`.
Currently it throws a `NullReferenceException` which causes problems.

We did discuss using an empty string in this case, but that may cause issues if something is trying to parse it into something else later on. In theory a `NULL` setting should now default to the default value of the setting (same as if it was not specified at all)

### DIFF
--- a/Source/Libraries/GSF.Core.Shared/StringExtensions.cs
+++ b/Source/Libraries/GSF.Core.Shared/StringExtensions.cs
@@ -358,7 +358,6 @@ namespace GSF
 
             foreach (string key in pairs.Keys)
             {
-               
                 value = pairs[key];
                 
                 if (value is null)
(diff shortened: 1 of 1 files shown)


Source/Libraries/GSF.Core.Shared/StringExtensions.cs
```diff
@@ -358,7 +358,11 @@ public static string JoinKeyValuePairs(this IDictionary<string, string> pairs, c
 
             foreach (string key in pairs.Keys)
             {
+
```
StephenCWills, line 361:
```suggestion
```

StephenCWills (changes requested):

StephenCWills (approved):

ritchiecarroll (approved):
Most use cases for this function are originally parsed from a string, so this would be unaffected. The only scenario I could see this being an issue would be for manually created dictionaries, perhaps like a TSL adapter that allowed a null string for a connection string property and was automatically parsing adapter connection string with the connection string parser... Regardless, looks safe enough.

PR #149, opened 2023-08-15
Added ConfigurationFrameVersion parameter to MPFP.

ritchiecarroll:
This allows specific config frame version to be requested from device overriding default behavior that otherwise requests latest configuration frame version for current protocol. This is useful in cases such as when IEEE C37.118-2011 is the selected protocol. The 2011 protocol selection defaults to requesting a configuration frame 3 frame, however, if remote device does not support this, this property can be set to send a configuration frame 2 frame instead. Property can be assigned as a MPFP connection string parameter, e.g.: `configurationFrameVersion=2` or `configurationFrameVersion=SendConfigurationFrame2`.


StephenCWills (approved):
Cool changes. 😎

PR #152, opened 2023-08-17
Use builder pattern to enable mapping of custom routes in WebPageController

StephenCWills:


Source/Libraries/GSF.Web/Hosting/WebPageController.cs
```diff
@@ -235,31 +286,75 @@ public IEnumerable<object> GetServices(Type serviceType)
...
+            webPageControllerConfig(builder);
+
             HttpConfiguration httpConfig = new HttpConfiguration();
-            httpConfig.DependencyResolver = GetDependencyResolver(webServer, model, modelType);
+            httpConfig.DependencyResolver = GetDependencyResolver(webServer, builder.Model, builder.ModelType);
+
+            AuthenticationOptions options = builder.AuthenticationOptions;
 
-            if (options != null)
+            if (!(options is null))
```
ritchiecarroll, line 345: FYI, the expression `if (options is not null)` should be supported now with latest C# version - all libraries in GSF have been updated to use this.

StephenCWills:
```suggestion
            if (options is not null)
```

ritchiecarroll (approved):
Looks good to me! Haven't tested any of this personally, but presume you have.

PR #156, opened 2023-08-29
Stat dups fix

ritchiecarroll:
Added a secondary check to prevent duplicated statistics from being created and updated statistics validation step in `CommonPhasorServices` to check for and remove duplicate statistics.

### DIFF
--- a/Source/Libraries/Adapters/PhasorProtocolAdapters/CommonPhasorServices.cs
+++ b/Source/Libraries/Adapters/PhasorProtocolAdapters/CommonPhasorServices.cs
@@ -1403,13 +1403,10 @@ namespace PhasorProtocolAdapters
                 const string QueryDuplicateStatistics = "SELECT SignalReference, COUNT(SignalReference) AS Occurrences FROM Measurement WHERE SignalTypeID = {0} GROUP BY SignalReference HAVING COUNT(SignalReference) > 1";
                 const string DeleteDuplicateStatistics = "DELETE FROM Measurement WHERE SignalReference = {0} AND SignalTypeID = {1} AND PointID NOT IN (SELECT MIN(PointID) FROM Measurement WHERE SignalReference = {0} AND SignalTypeID = {1})";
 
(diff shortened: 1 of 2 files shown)


Source/Libraries/Adapters/PhasorProtocolAdapters/CommonPhasorServices.cs
```diff
@@ -1390,6 +1397,35 @@ private static void ValidateStatistics(AdoDataConnection database, Action<string
                     database.Connection.ExecuteNonQuery(string.Format(OutputStreamStatInsertFormat, signalIndex, statName, statDescription, statMethodSuffix, statType, statFormat, signalIndex == 11 ? 1 : 0, statLoadOrder));
                 }
             }
+
+            try
+            {
+                const string QueryDuplicateStatistics = "SELECT SignalReference, COUNT(SignalReference) AS Occurrences FROM Measurement WHERE SignalTypeID = {0} GROUP BY SignalReference HAVING COUNT(SignalReference) > 1";
+                const string DeleteDuplicateStatistics = "DELETE FROM Measurement WHERE SignalReference = {0} AND SignalTypeID = {1} AND PointID NOT IN (SELECT MIN(PointID) FROM Measurement WHERE SignalReference = {0} AND SignalTypeID = {1})";
```
StephenCWills, line 1404: Looks like you're using `ExecuteNonQuery<TConnection>(this TConnection connection, ...)` which doesn't accept these format parameters. This `ValidateStatistics()` method uses `string.Format()` for a lot of its queries.

StephenCWills (changes requested):

Source/Libraries/Adapters/PhasorProtocolAdapters/CommonPhasorServices.cs
ritchiecarroll: Good catch. Not sure why I was referencing `.Connection` there instead of using `AdoDataConnection` wrappers directly - copy/paste I guess, I so rarely use those older functions...

clackner-gpa (approved):

StephenCWills (approved):

PR #163, opened 2023-09-14
Fixed SQL Exception when using Pageination and Filters

clackner-gpa:
This fixes a bug where we modified the `SearchText` properties of a Filter when building the where clause.


StephenCWills (approved):
Nothing wrong with this PR, but this function needs to be redone to avoid SQL injection. Would also be nice if some consideration was made for other database engines besides SQL Server.

PR #169, opened 2023-09-28
Added feature to only export on downsampling interval for CsvExportAdapter

ritchiecarroll:
When using normalized sub-second downsampling in the CsvExportAdapter, only export measurements that fall on the downsampling interval.

This corrects issue identified in ticket #1518

### DIFF
--- a/Source/Libraries/Adapters/CsvAdapters/CsvExportAdapter.cs
+++ b/Source/Libraries/Adapters/CsvAdapters/CsvExportAdapter.cs
@@ -100,6 +100,8 @@ namespace CsvAdapters
         private readonly object m_activeFileLock;
         private readonly ScheduleManager m_scheduleManager;
         private readonly Dictionary<Guid, long> m_lastTimestamps;
+        private double m_downsampleInterval;
+        private int m_downsampleMilliseconds;
         private long m_totalExports;
         private bool m_disposed;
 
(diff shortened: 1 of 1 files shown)


Source/Libraries/Adapters/CsvAdapters/CsvExportAdapter.cs
```diff
@@ -356,10 +375,21 @@ public override void QueueMeasurementsForProcessing(IEnumerable<IMeasurement> me
...
+                            continue;
+
+                        // If last measurement timestamp is within the downsampling interval, then we have
+                        // a measurement that is too close to the last measurement and should be ignored
+                        if ((timeNormalizedMeasurement.Timestamp - lastTimestamp).ToSeconds() < DownsampleInterval)
+                            continue;
+                    }
+
+                    // For normalized sub-second downsampling, only export measurements that fall on the downsampling interval
+                    if (FramesPerSecond > 0 && m_downsampleMilliseconds < 1000 & (int)timeNormalizedMeasurement.Timestamp.DistanceBeyondSecond().ToMilliseconds() % m_downsampleMilliseconds != 0)
```
StephenCWills, line 392: Should this be using `&&` instead of `&`?

StephenCWills: It seems like this logic wouldn't work for some intervals. For example, a downsample interval of 0.667 would end up behaving like a downsample interval of 1 where measurements are only exported at 667 milliseconds past the top of the second.

I think maybe what you were going for would be to compute a frame index that is equivalent to the downsampling interval at the specified framerate. Then, from a measurement timestamp, compute the number of frames since some epoch and determine if that is divisible by the frame index.

PR #171, opened 2023-09-28
Fixes for downsampling in CsvExportAdapter

StephenCWills:
* This frame-counting strategy works for both subsecond intervals and long intervals so I just went all in and got rid of the special casing for `DownsampleInterval < 1.0D`.
* I changed the terminology from "normalized" to "aligned" because I felt it was easier to understand.
* Although it would probably be due to a configuration error, I limited `m_downsampleFrameCount` to be at least 1 to prevent divide-by-zero errors.
* I tweaked the condition that checks downsample interval against the timestamp difference because it's unnecessary when performing alignment and would only invite potential issues due to rounding errors.


ritchiecarroll (approved):
Thanks Stephen, looks good.

PR #175, opened 2023-10-24
Model controller update

clackner-gpa:
Temporary change to rework TSX component to remove SQL Injections.


StephenCWills (commented):
I'll need to do a cleanup before approving, but I had a couple questions about how this is supposed to work.

Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -498,15 +501,26 @@ public virtual IHttpActionResult GetPagedList([FromBody] PostData postData, int
...
+            foreach (Search search in searches)
+            {
+                nParameter = parameters.Count();
                 bool isQuery = false;
+                string query = "";
                 string searchText = search.SearchText;
                 if (searchText == string.Empty) searchText = "%";
                 else searchText = searchText.Replace("*", "%");
 
+                if (Array.IndexOf(new[] { "=", "<>", "<", ">", "IN" }, search.Operator) < 0)
```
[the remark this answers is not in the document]
clackner-gpa: good catch. `LIKE` and `NOT LIKE` are missing from this list. I think they should be added.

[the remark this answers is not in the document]
StephenCWills: Argh, I meant to say `NOT IN`, but I guess `NOT LIKE` may also be important. I'll get them added.

Source/Libraries/GSF.Core/Data/Model/SQLSearchFilter.cs
```diff
@@ -0,0 +1,176 @@
...
+             * of the above to avoid legacy issues
+             */
+
+            if (Type != "query" && (Operator == "IN" || Operator == "NOT IN"))
+            {
+                IEnumerable<string> valueList = searchText
+                    .Replace("(", "")
+                    .Replace(")", "")
+                    .Split(',')
+                    .Select(t => $"'{t}'");
```
StephenCWills, line 159: So the syntax for IN/NOT IN is going to be something like `AN,BN,CN`? Does the user have to enter the commas? Should we trim each element in the list? What about values that contain commas or parens; do we not care?

Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -517,37 +507,23 @@ public virtual IHttpActionResult GetPagedList([FromBody] PostData postData, int
...
 
-            string whereClause = string.Join(" AND ", searches.Select(search => {
-                bool isQuery = false;
-                string searchText = search.SearchText;
-                if (searchText == string.Empty) searchText = "%";
-                else searchText = searchText.Replace("*", "%");
+            foreach (SQLSearchFilter search in searches)
+            {
+                if (search.SearchText == string.Empty)
+                    search.SearchText = "%";
```
StephenCWills, line 517: Perhaps this only makes sense when the operator is LIKE/NOT LIKE?

Also, the original logic supported asterisks. Are we doing away with that?

Source/Libraries/GSF.Core/Data/Model/SQLSearchFilter.cs
clackner-gpa: No the Commas get added by the UI. These are primary for the Filters that show Checkboxes. We need this for legacy reasons. This is currently working and we want to avoid breaking anything for the time being so trimming or considering values with commas may cause issues.

Source/Libraries/GSF.Web/Model/ModelController.cs
clackner-gpa: Good Catch it's missing `search.SearchText= search.SearchText.Replace("*", "%");

StephenCWills: Then what do you think of this? or do we need to keep the old logic and ignore the operator?

```suggestion
                if (search.Operator == "LIKE" || search.Operator == "NOT LIKE")
                {
                    if (search.SearchText == string.Empty)
                        search.SearchText = "%";
                    else
                        search.SearchText = search.SearchText.Replace("*", "%");
                }
```

StephenCWills (approved):

PR #178, opened 2023-11-29
View only field

gcsantos-gpa:
Addition of view only fields for a couple of reasons:
   1. It's cumbersome to add custom endpoints for custom view that just pull in slightly more information on model controllers.
   2. There is a legitimate use-case for fields that don't make sense to be updated within openXDA, such as primary keys. The primary key use-case is covered, but the use case of similar generated fields is not.


gcsantos-gpa:
Chasing down a bug this introduced....

PR #179, opened 2023-11-29
Edit view split

gcsantos-gpa:


Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -97,10 +108,20 @@ public ModelController()
...
                 SecurityType = "Claims";
 
-                IEnumerable<ClaimAttribute> claimAttributes = typeof(T).GetCustomAttributes<ClaimAttribute>();
+                IEnumerable<ClaimAttribute> claimViewAttributes = typeof(T).GetCustomAttributes<ClaimAttribute>();
+                IEnumerable<ClaimAttribute> claimEditAttributes = typeof(U).GetCustomAttributes<ClaimAttribute>();
 
-                foreach (ClaimAttribute claimAttribute in claimAttributes)
+                foreach (ClaimAttribute claimAttribute in claimViewAttributes)
+                {
+                    if (claimAttribute.Verb != "GET") continue;
```
clackner-gpa, line 116: This does not work the way you intended. The `SearchableList` endpoint for instance is `POST` but should be using the `claimViewAttributes`.

Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -43,7 +44,16 @@ namespace GSF.Web.Model
...
+    public class ModelController<T> : ModelController<T, T> where T : class, new() { }
+
+    /// <summary>
+    /// Base Class for A Webcontroller that provides common Model endpoints such as Edit, Delete, Search.
+    /// </summary>
+    /// <typeparam name="T">The corresponding Model for Search/Fetch Results.</typeparam>
+    /// <typeparam name="U">The corresponding Model for database editing.</typeparam>
+    public class ModelController<T,U> : ApiController 
+        where T : class, new() 
+        where U : class, new()
```
clackner-gpa, line 56: would it be worth further restricting U to be `U: T, new()`?
The use case would be to only allow this for Views that include the entire Table.
That would also allow you to use a single `PrimaryKeyField`

clackner-gpa (dismissed):

Source/Libraries/GSF.Web/Model/ModelController.cs
gcsantos-gpa: As per latest commit, T is now set to be a child of U.

StephenCWills (changes requested):

Source/Libraries/GSF.Web/Model/ModelController.cs
```diff
@@ -121,18 +141,18 @@ public ModelController()
                 SecurityType = "Roles";
                 PostRoles = typeof(T).GetCustomAttribute<PostRolesAttribute>()?.Roles ?? "Administrator";
```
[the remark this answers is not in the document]
gcsantos-gpa: Good catch

StephenCWills (approved):

PR #188, opened 2023-12-06
Fixed Azure query to match email instead of ID

clackner-gpa:
Changed ADOSecurity Provider to use email field to match users instead of ID. That way we don't have to add `#Ext#` anywhere


ritchiecarroll (approved):
I presume `Username` in these cases is always an e-mail. I think this update is fine so long as we always login into Azure AD with an e-mail - and since I "think" this is true, this should work.

Originally, I presumed using the `Users` dictionary might be faster than a manual query - but it may just be the same filter on the backend anyway -- and you can't really test without access to a huge AD instance...

So if this query is not slow on a huge AD system, I think this update looks good.

PR #193, opened 2023-12-20
Improved UI feedback on out of sync operations for PDCs

ritchiecarroll:
This update places the UI feedback at the proper level in the hierarchy for PDCs

### DIFF
--- a/Source/Libraries/GSF.PhasorProtocols/UI/DataModels/RealTimeStream.cs
+++ b/Source/Libraries/GSF.PhasorProtocols/UI/DataModels/RealTimeStream.cs
@@ -97,15 +97,6 @@ namespace GSF.PhasorProtocols.UI.DataModels
             }
         }
 
-        ///// <summary>
-        ///// Gets flag that determines if the <see cref="RealTimeStream"/> is a concentrator device.
-        ///// </summary>
-        //public bool IsConcentratorDevice
-        //{
-        //    get => !Acronym.Equals("DIRECT CONNECTED");
-        //    set { }
-        //}
-
         /// <summary>
         /// Gets or sets the name of the current <see cref="RealTimeStream"/>.
         /// </summary>
--- a/Source/Libraries/GSF.PhasorProtocols/UI/DataModels/RealTimeStream.cs
+++ b/Source/Libraries/GSF.PhasorProtocols/UI/DataModels/RealTimeStream.cs
@@ -52,7 +52,7 @@ namespace GSF.PhasorProtocols.UI.DataModels
         private bool m_enabled;
         private bool m_expanded;
         private string m_statusColor;
-        private bool m_configurationOutOfSync;
+        private bool? m_configurationOutOfSync;
         private ObservableCollection<RealTimeDevice> m_deviceList;
 
         #endregion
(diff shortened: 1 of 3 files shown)


Source/Libraries/GSF.PhasorProtocols/UI/DataModels/RealTimeStream.cs
```diff
@@ -154,14 +157,62 @@ public string StatusColor
...
+        /// Gets content message for input stream configuration sync state.
+        /// </summary>
+        public string ConfigurationOutOfSyncContent
+        {
+            get => DeviceStateUnavailable ? "Loading..." :
+                m_configurationOutOfSync is null ? "Checking Config..." :
+                    m_configurationOutOfSync.Value ?
+                        "Config Out of Sync" :
+                        "Config In Sync";
+            set { }
```
StephenCWills, line 181: I don't get it. Why does it have a setter if the setter isn't used and doesn't do anything?

ritchiecarroll: WPF thing - it complained when it didn't exist on binding? Maybe I can do something in WPF to make it work with a readonly property - but this worked - and this was how other properties were doing it...

StephenCWills: Hm, I do see it in `RealTimeDevice`, but all other read-only bindings throughout the codebase just do `Target={Binding Prop}` instead of `Target={tsfBinding:Column Path=Prop}`.

ritchiecarroll: Honestly, I am not even sure what the difference is between those two targets except that we must have "modified" one to be custom. WPF and I don't get along well. This update too way many more hours than it should have fighting little things. That said, happy to try a different way if think it will make a difference. My thought is that we can eventually abandon this code base - that `RealTimeStream` class is way too complex. Maybe it needs to be, but it so, its too much jammed into that one file...

StephenCWills: `tsfBinding:Column` just sets a bunch of properties on the `Binding` class. The most relevant one is probably `Mode=TwoWay`.

https://github.com/GridProtectionAlliance/gsf/blob/cd87678635087a8c15ad9aa4f791cb9478e9c58d/Source/Libraries/GSF.TimeSeries/UI/Binding/Column.cs#L33-L52

Source/Libraries/GSF.TimeSeries/UI/Converters/ObjectToVisibilityConverter.cs
```diff
@@ -47,35 +45,24 @@ public class ObjectToVisibilityConverter : IValueConverter
...
-                    {
-                        if (temp > 0)
-                            value = true;
-                        else
-                            value = false;
-                    }
-                }
+                if (int.TryParse(value.ToString(), out int intVal))
+                    value = intVal > 0;
+            }
```
StephenCWills, line 54: The parsing step seems pointless. Isn't this viable?
```c#
if (value is int intVal)
    value = intVal > 0;
```

Source/Libraries/GSF.PhasorProtocols/UI/DataModels/RealTimeStream.cs
ritchiecarroll: Yeah, guess that would be why `set` is required.

StephenCWills: And, by extension, why it's the wrong type of binding for read-only properties. But I suppose I will drop it since I don't really care enough to fix it myself either.

StephenCWills (approved):

PR #202, opened 2024-01-18
Updated SQLSearchModifier to return string

clackner-gpa:
I realized the whole point of the `SQLSearchModifier` is to generate complex conditions that won't work inside a `SQLSearch`object with SQL injection avoidance.


StephenCWills (approved):

StephenCWills:
Rebased in order to sign commits.

PR #204, opened 2024-01-21
FileAdapters: Updates to ProcessLauncher 

ritchiecarroll:
ProcessLauncher adapter updated to terminate child processes when a `Kill` command is invoked.

Note that when launching newer Grafana instances, the `grafana-server` launches a child process called `grafana`.

When process launcher is disposed, it properly handles terminations of child processes by using the `ChildProcessManager` internally, at least when the property `ForceKillOnDispose` is `true` (the default value). However, when `Kill` method was invoked from the CLI, it simply killed the parent process, leaving the child process intact, causing some undesired/unexpected behavior when the adapter was re-initialized.

This PR causes the `Kill` to also dispose the `ChildProcessManager`, thus terminating all child processes as well.

The common use case is: invoke `Kill` to stop Grafana, update Grafana config (updating plugins, database etc.), then re-initialize Grafana process adapter.

Note that the diffs are hard to compare due to some simple code cleanup, but here are the relevant updates:

```c#
    /// <summary>
    /// Releases the unmanaged resources used by the <see cref="ProcessLauncher"/> object and optionally releases the managed resources.
    /// </summary>
    /// <param name="disposing"><c>true</c> to release both managed and unmanaged resources; <c>false</c> to release only unmanaged resources.</param>
    protected override void Dispose(bool disposing)
    {
        if (m_disposed)
            return;

        try
        {
            if (!disposing)
                return;

            if (ForceKillOnDispose)
                Kill();

            m_process.Dispose();
            m_process.OutputDataReceived -= ProcessOutputDataReceived;
            m_process.ErrorDataReceived -= ProcessErrorDataReceived;

            m_processUtilizationCalculator.Dispose();
        }
        finally
        {
            m_disposed = true;          // Prevent duplicate dispose.
            base.Dispose(disposing);    // Call base class Dispose().
        }
    }

    /// <summary>
    /// Stops the launched process.
    /// </summary>
    [AdapterCommand("Stops the launched process.")]
    public void Kill()
    {
        try
        {
            m_process.Kill();

            // Kill any child processes that were launched by the process by disposing of the child process manager
            m_childProcessManager?.Dispose();
        }
        catch (Exception ex)
        {
            OnProcessException(MessageLevel.Warning, new InvalidOperationException($"Exception encountered while attempting to stop process launched from \"{FileName}\": {ex.Message}", ex));
        }
    }
```


Source/Libraries/Adapters/FileAdapters/ProcessLauncher.cs
```diff
@@ -38,953 +38,902 @@
...
+            throw new ArgumentException($"Cannot launch process: required \"{nameof(FileName)}\" parameter is missing from connection string.");
+        }
 
-            // Load optional parameters
-            if (settings.TryGetValue(nameof(SupportsTemporalProcessing), out setting))
-                m_supportsTemporalProcessing = setting.ParseBoolean();
-            else
-                m_supportsTemporalProcessing = DefaultSupportsTemporalProcessing;
+        // Load optional parameters
+        m_supportsTemporalProcessing = settings.TryGetValue(nameof(SupportsTemporalProcessing), out setting) && setting.ParseBoolean();
```
StephenCWills, line 613: I get that these are equivalent because `DefaultSupportsTemporalProcessing` is `false`, but it still seems wrong to not use it here.
```suggestion
        m_supportsTemporalProcessing = settings.TryGetValue(nameof(SupportsTemporalProcessing), out setting) ? setting.ParseBoolean() : DefaultSupportsTemporalProcessing;
```

StephenCWills (approved):

Source/Libraries/Adapters/FileAdapters/ProcessLauncher.cs
```diff
@@ -38,953 +38,902 @@
...
+
+    /// <summary>
+    /// Default value for the <see cref="CreateNoWindow"/> property.
+    /// </summary>
+    public const bool DefaultCreateNoWindow = true;
+
+    /// <summary>
+    /// Default value for the <see cref="WindowStyle"/> property.
+    /// </summary>
+    public const string DefaultWindowStyle = nameof(WindowStyle.Normal);
```
StephenCWills, line 86:
```suggestion
    public const string DefaultWindowStyle = nameof(ProcessWindowStyle.Normal);
```

PR #207, opened 2024-01-26
GSF-4 TSFManager localization fixes

StephenCWills:
The first commit applies the system's current culture settings to the WPF UI. Without this, it will default to en-US, which explains why numbers were displaying with periods instead of commas when using `en-AT` culture.

The second commit relaxes the regex validation for lat/long so that a number like `-98,6` will not fail validation. The regex was being applied to a string created using the current culture settings regardless of what was being used for the display. This explains why the validation was failing despite the UI showing a number with a valid format.

The first commit will need to be applied to every Manager application individually, not just TSFManager.

### DIFF
--- a/Source/Applications/TsfManager/App.xaml.cs
+++ b/Source/Applications/TsfManager/App.xaml.cs
@@ -24,11 +24,13 @@
 //******************************************************************************************************
 
 using System;
+using System.Globalization;
 using System.Security.Principal;
 using System.Windows;
-using GSF.Windows.ErrorManagement;
+using System.Windows.Markup;
 using GSF.Reflection;
 using GSF.TimeSeries.UI;
+using GSF.Windows.ErrorManagement;
 
 namespace TsfManager
 {
@@ -54,6 +56,7 @@ namespace TsfManager
         public App()
         {
             AppDomain.CurrentDomain.SetPrincipalPolicy(PrincipalPolicy.WindowsPrincipal);
+            Current.Startup += Application_Startup;
 
             m_errorLogger = new ErrorLogger
             {
@@ -97,6 +100,12 @@ namespace TsfManager
         #endregion
 
         #region [ Methods ]
+
+        private void Application_Startup(object sender, StartupEventArgs e)
+        {
+            XmlLanguage xmlLanguage = XmlLanguage.GetLanguage(CultureInfo.CurrentCulture.IetfLanguageTag);
+            FrameworkElement.LanguageProperty.OverrideMetadata(typeof(FrameworkElement), new FrameworkPropertyMetadata(xmlLanguage));
+        }
 
         private string ErrorText()
         {
--- a/Source/Applications/TsfManager/App.xaml.cs
+++ b/Source/Applications/TsfManager/App.xaml.cs
@@ -28,6 +28,7 @@ using System.Globalization;
 using System.Security.Principal;
 using System.Windows;
 using System.Windows.Markup;
+using GSF.Diagnostics;
 using GSF.Reflection;
 using GSF.TimeSeries.UI;
 using GSF.Windows.ErrorManagement;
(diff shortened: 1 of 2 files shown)


Source/Applications/TsfManager/App.xaml.cs
```diff
@@ -98,6 +101,12 @@ public Guid NodeID
 
         #region [ Methods ]
 
+        private void Application_Startup(object sender, StartupEventArgs e)
+        {
+            XmlLanguage xmlLanguage = XmlLanguage.GetLanguage(CultureInfo.CurrentCulture.IetfLanguageTag);
```
ritchiecarroll, line 106: For distribution to other TSL applications: I haven't looked at what exceptions these functions calls might throw, but should we wrap these updates in a `try / catch` then just log exception, e.g., with `Logger.SwallowException`.  Could be safe enough, but this is a critical startup area.

Source/Applications/TsfManager/App.xaml.cs
```diff
@@ -98,6 +101,12 @@ public Guid NodeID
 
         #region [ Methods ]
 
+        private void Application_Startup(object sender, StartupEventArgs e)
```
ritchiecarroll, line 104: I think this is just the test app in GSF, so I presume if these are "accepted" changes, the next step would be distribution of this code to openPDC, openHistorian, SIEGate and ProjectAlpha at a minimum?

ritchiecarroll (commented):
These are just some questions and comments...

Source/Applications/TsfManager/App.xaml.cs
ritchiecarroll: You answered this question in PR notes! :-)

Source/Applications/TsfManager/App.xaml.cs
StephenCWills: Right, so looking into this, I first found this note about `IetfLanguageTag`.

> This property (and the [GetCultureInfoByIetfLanguageTag](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.getcultureinfobyietflanguagetag?view=netframework-4.8) method) is deprecated. Instead, you should use the [CultureInfo.Name](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.cultureinfo.name?view=netframework-4.8) property. IETF tags and names are identical.

Neither `IetfLanguageTag` nor `CultureInfo.Name` have any documented exceptions. `XmlLanguage.GetLanguage()` does throw `ArgumentException` if the input does not conform to the syntax specified in RFC 3066. The documentation for `IetfLanguageTag` and `CultureInfo.Name` does not mention RFC 3066 so it's perhaps reasonable to assume something could go wrong in an edge case.

That said, I also tried this in PowerShell and didn't get any errors.

```powershell
Add-Type -AssemblyName PresentationCore
[Globalization.CultureInfo]::GetCultures("AllCultures") | % { [Windows.Markup.XmlLanguage]::GetLanguage($_.Name) }
```

As for `DependencyProperty.OverrideMetadata()`, it doesn't look like we should need to worry about exceptions there as it relates to overriding metadata on read-only properties and properties where metadata was "already established".

Anyway, you bring up a good point about this being a critical startup area. Even if there was provably no possibility for exceptions here, it would still be a good idea to add the try-catch.

Source/Applications/TsfManager/App.xaml.cs
StephenCWills: Yes. I was also thinking of substationSBG and openMIC.

ritchiecarroll (commented):
Looks good.

StephenCWills:
Rebased to get up-to-date with master.

Source/Libraries/GSF.PhasorProtocols/UI/DataModels/Device.cs
```diff
@@ -314,7 +314,7 @@ public int AccessID
         /// Gets or sets <see cref="Device"/> Longitude.
         /// </summary>
         // Because of database design, no validation attributes are applied.        
-        [RegularExpression(@"^[-]?([0-9]{1,3})?([.][0-9]{1,6})?$", ErrorMessage = "Invalid value. Please provide value in decimal(9,6) format.")]
+        [RegularExpression(@"^[-]?([0-9]{1,3})?([.,][0-9]{1,6})?$", ErrorMessage = "Invalid value. Please provide value in decimal(9,6) format.")]
```
clackner-gpa, line 317: I am assuming `decimal(9,6)` format is `###./,######`? Can we rephrase that to make this clear in the UI?

clackner-gpa (approved):

Source/Libraries/GSF.PhasorProtocols/UI/DataModels/Device.cs
StephenCWills: It comes from how precision and scale for a data type are specified in SQL.
https://learn.microsoft.com/en-us/sql/t-sql/data-types/decimal-and-numeric-transact-sql#arguments

Perhaps it was described that way to be precise, because it's actually the data type we use in the database.

https://github.com/GridProtectionAlliance/gsf/blob/b4a75117714335053892e42c8caf1aaaa9297da4/Source/Libraries/GSF.TimeSeries/Data/SQL%20Server/GSFSchema.sql#L326-L327

But I can understand why this isn't really a great way to describe the syntax to the user. Perhaps the most intuitive way to describe it would be between -1000 and 1000 with up to six decimal places. Though it might be better to specify -180 to 180 for longitude and -90 to 90 for latitude, even if that doesn't really match the regex or the data type.

PR #210, opened 2024-01-31
Ignore case since db's ussually aren't case sensitive

gcsantos-gpa:


Source/Libraries/GSF.Core/Data/Model/SQLSearchFilter.cs
```diff
@@ -95,11 +95,12 @@ public string Operator
         public string Type
         {
             get => m_type;
-            set
+            set //
```
gcsantos-gpa, line 98:
```suggestion
            set
```

clackner-gpa (approved):

PR #220, opened 2024-02-04
Added Delayed Restart Option

clackner-gpa:
CommandLine takes `restartDelay=15` which means wait 15 seconds and then restarts service.


ritchiecarroll (changes requested):
Looks good, however, I suggest renaming parameter to `restartWithDelay` otherwise `restartDelay` sounds like a sub-ordinate parameter to `restart` instead of an optional secondary mode.

ritchiecarroll (approved):

PR #223, opened 2024-02-12
Added grafana Switch

clackner-gpa:
Added Switch Function.
It takes a slice and uses the first number as an index to select a value from all other values in the slice.


ritchiecarroll (approved):
I posted an update here

PR #253, opened 2024-06-10
GSF.Web: Add an option for web applications to redirect all 404 Not Found to a single page

StephenCWills:
Adds an option to `IWebPageControllerBuilder` to redirect all responses that would return 404 Not Found to the Default Web Page instead.

This is done by registering a branching middleware before the `WebPageController` middleware on the main branch of the Owin pipeline. The branching middleware invokes the `WebPageController` middleware on the main branch, and if the response from the main branch is 404 Not Found, then it subsequently invokes a second `WebPageController` on a separate branch of the pipeline. The second `WebPageController` uses the same `WebServer` instance as the first, but has a slightly modified route template to force all resources to resolve to the default web page.

Since the second `WebPageController` is only invoked when the first responds with 404 Not Found, this will not mask other errors such as 500 Internal Server Error. And since the `WebPageController` maps all possible routes, this should only redirect requests that fail to locate resources on the following lines of code:

Embedded/static resource:
https://github.com/GridProtectionAlliance/gsf/blob/6720c394fbc7da0e45a9df84e7922cf88ac37a35/Source/Libraries/GSF.Web/Hosting/WebServer.cs#L225

Cached resource:
https://github.com/GridProtectionAlliance/gsf/blob/6720c394fbc7da0e45a9df84e7922cf88ac37a35/Source/Libraries/GSF.Web/Hosting/WebServer.cs#L263

However, it may be worth noting that if a custom route is added via `IWebPageControllerBuilder.UseCustomRoutes()`, and that route attempts to explicitly respond with 404 Not Found, this new setting would redirect it.

### DIFF
--- a/Source/Libraries/GSF.Web/Hosting/WebPageController.cs
+++ b/Source/Libraries/GSF.Web/Hosting/WebPageController.cs
@@ -24,6 +24,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
+using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


clackner-gpa (approved):

clackner-gpa:
Added @ritchiecarroll for visibility. If no one has any comments I will merge tomorrow.

prestoncraw (approved):

PR #256, opened 2024-06-14
GSF.COMTRADE: Use invariant culture for all numeric conversions

StephenCWills:
Fix culture issues reported in https://github.com/gemstone/comtrade/issues/2

### DIFF
--- a/Source/Libraries/GSF.COMTRADE/SampleRate.cs
+++ b/Source/Libraries/GSF.COMTRADE/SampleRate.cs
@@ -64,8 +64,8 @@ namespace GSF.COMTRADE
             if (parts.Length < 2 || (!useRelaxedValidation && parts.Length != 2))
                 throw new InvalidOperationException($"Unexpected number of line image elements for sample rate definition: {parts.Length} - expected 2{Environment.NewLine}Image = {lineImage}");
 
-            Rate = double.Parse(parts[0].Trim());
-            EndSample = long.Parse(parts[1].Trim());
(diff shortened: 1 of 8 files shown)


Source/Libraries/GSF.COMTRADE/SampleRate.cs
```diff
@@ -75,11 +75,16 @@ public SampleRate(string lineImage, bool useRelaxedValidation = false)
...
-        }
+        public override string ToString() => // samp,endsamp
+            Format($"{Rate},{EndSample}");
+
+        #endregion
+
+        #region [ Static ]
+
+        // Static Methods
+        private static string Format(FormattableString formattableString) =>
```
ritchiecarroll, line 86: Huh - this is a new on me. Odd that an interpolated string can change its stripes in what looks to be post interpolation...

ritchiecarroll (approved):

Source/Libraries/GSF.COMTRADE/SampleRate.cs
StephenCWills: Yeah, the compiler has to be interpreting the string differently depending on how it's used. Intellisense still reports the interpolated string as `System.String` on mouseover, but allows me to assign it to a variable of type `FormattableString`. It makes me a bit feel like I'm doing something wrong, but the compiler also doesn't let me assign any old string to a variable of type `FormattableString` so it must be fine so long as there's no compiler errors.

PR #260, opened 2024-06-23
Added Logic to prevent excessive triggering if condition is met

clackner-gpa:
This adds a Connectionstring parameter to prevent an email from going out on every recalculation as long as the condition is met.
If `MultiTriggerPrevention` is `true` an email will only go out the first time the condition evaluates to `true` subsequent evaluation will be ignored until the condition evaluates to `false` at least once.

This prevents the system from sending out continous emails when using complex alarm logic

### DIFF
--- a/Source/Libraries/Adapters/DynamicCalculator/EmailNotifier.cs
+++ b/Source/Libraries/Adapters/DynamicCalculator/EmailNotifier.cs
@@ -46,12 +46,14 @@ namespace DynamicCalculator
         private const int DefaultFramesPerSecond = 30;
         private const double DefaultLagTime = 5.0D;
         private const double DefaultLeadTime = 5.0D;
+        private const bool DefaultSpamTriggerPrevention = false;
 
         // Fields
         private readonly Mail m_mailClient;
         private long m_expressionSuccesses;
         private long m_expressionFailures;
         private long m_totalEmailOperations;
+        private bool m_multiTriggerPrevention = false;
 
         #endregion
 
@@ -67,6 +69,17 @@ namespace DynamicCalculator
 
         #region [ Properties ]
 
+        /// <summary>
+        /// Gets or sets the Double Trigger prevention.
+        /// </summary>
+        [ConnectionStringParameter]
+        [Description("Define whether the Trigger continously being met will send multiple email or if this needs to be prevented.")]
+        [DefaultValue(DefaultSpamTriggerPrevention)]
+        public bool MultiTriggerPrevention
+        {
+            get;
+            set;
+        }
         /// <summary>
         /// Gets or sets the textual representation of the boolean expression.
         /// </summary>
@@ -376,6 +389,10 @@ namespace DynamicCalculator
 
             if (settings.TryGetValue(nameof(EnableSSL), out setting) && !string.IsNullOrWhiteSpace(setting))
                 EnableSSL = setting.ParseBoolean();
+            
+            if (settings.TryGetValue(nameof(MultiTriggerPrevention), out setting) && !string.IsNullOrWhiteSpace(setting))
+                MultiTriggerPrevention = setting.ParseBoolean();
+            m_multiTriggerPrevention = false;
         }
 
         /// <summary>
@@ -386,6 +403,9 @@ namespace DynamicCalculator
         {            
             if (value.ToString().ParseBoolean())
             {
+                if (m_multiTriggerPrevention && MultiTriggerPrevention)
+                    return;
+                m_multiTriggerPrevention = true;
                 m_expressionSuccesses++;
                 m_mailClient.Body = Body.Interpolate(Variables);
                 m_mailClient.Send();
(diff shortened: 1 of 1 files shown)


Source/Libraries/Adapters/DynamicCalculator/EmailNotifier.cs
```diff
@@ -46,12 +46,14 @@ public class EmailNotifier : DynamicCalculator
...
         private const double DefaultLagTime = 5.0D;
         private const double DefaultLeadTime = 5.0D;
+        private const bool DefaultSpamTriggerPrevention = false;
 
         // Fields
         private readonly Mail m_mailClient;
         private long m_expressionSuccesses;
         private long m_expressionFailures;
         private long m_totalEmailOperations;
+        private bool m_multiTriggerPrevention = false;
```
ritchiecarroll, line 56: The `MultiTriggerPrevention` property is an auto-property. By convention, if this property had a backing field it would be named `m_multiTriggerPrevention`. The `m_multiTriggerPrevention` field is instead being used a state that detects "when" multiple triggers have occurred. Without carefully looking, one might assume that the `m_multiTriggerPrevention` was a backing field, when it is not. I suggest renaming the field to something like `m_triggerDetected`. 

Also, assigning a field to its default value (`false`) is unnecessary.

Source/Libraries/Adapters/DynamicCalculator/EmailNotifier.cs
```diff
@@ -67,6 +69,17 @@ public class EmailNotifier : DynamicCalculator
 
         #region [ Properties ]
 
+        /// <summary>
+        /// Gets or sets the Double Trigger prevention.
+        /// </summary>
+        [ConnectionStringParameter]
+        [Description("Define whether the Trigger continously being met will send multiple email or if this needs to be prevented.")]
```
ritchiecarroll, line 76: It is possible that this property, and associated functionality, may be more appropriately placed in the base class `DynamicCalculator`. The `DynamicCalculator` provides common functionality for itself, the `EmailNotifier`, the `DatabaseNotifier` and the `DyanmicFilter`.  That is, would basic calculations and associated derivatives, all benefit from the ability to suppress multiple triggers, or do you think this more of a specific functionality that only affects triggered e-mails? 

FYI, this is just a thought, I do not have a strong opinion here.
 
If this property gets moved into `DynamicCalculator`, the description should be updated to be more abstract, i.e., not specifically about e-mails.

Also, correct spelling of text in `Description` attribute:
```
[Description("Defines flag that determines if the trigger continuously being met will send multiple email or should be prevented."]
```

Source/Libraries/Adapters/DynamicCalculator/EmailNotifier.cs
```diff
@@ -67,6 +69,17 @@ public class EmailNotifier : DynamicCalculator
 
         #region [ Properties ]
 
+        /// <summary>
```
ritchiecarroll, line 72: This property description text should match `Description` attribute text except for the "Gets or sets" prefix, so:

```
Gets or sets flag that determines if the trigger continuously being met will send multiple email or should be prevented.
```

Source/Libraries/Adapters/DynamicCalculator/EmailNotifier.cs
```diff
@@ -376,6 +389,10 @@ public override void Initialize()
 
             if (settings.TryGetValue(nameof(EnableSSL), out setting) && !string.IsNullOrWhiteSpace(setting))
                 EnableSSL = setting.ParseBoolean();
+            
+            if (settings.TryGetValue(nameof(MultiTriggerPrevention), out setting) && !string.IsNullOrWhiteSpace(setting))
+                MultiTriggerPrevention = setting.ParseBoolean();
+            m_multiTriggerPrevention = false;
```
ritchiecarroll, line 395: Suggest a new line between property initialization and field assignment.

Source/Libraries/Adapters/DynamicCalculator/EmailNotifier.cs
```diff
@@ -386,13 +403,17 @@ protected override void HandleCalculatedValue(object value)
         {            
             if (value.ToString().ParseBoolean())
             {
+                if (m_multiTriggerPrevention && MultiTriggerPrevention)
+                    return;
+                m_multiTriggerPrevention = true;
```
ritchiecarroll, line 408: Suggest a new line between `return` statement and field assignment.

ritchiecarroll (changes requested):
A few very minor updates. I leave the consideration on moving the functionality to the base class (as discussed below) to your discretion or postponement.

clackner-gpa:
The `DynamicCalculator` base class is not neccesarrily evaluating a `bool` so that won't work. Cleanest solution would bve an intermediate base class `DynamicTrigger` that uses `Dynamic Calculator` to trigger something and stick it in there. But that's too much work for now.

ritchiecarroll (approved):

PR #261, opened 2024-06-24
Update to Wizard screen for "Use Source Prefix" option

ritchiecarroll:
Fixes "Violation of UNIQUE KEY constraint 'UK_UniqueID'" for config frames that contain existing device acronyms and user selects "Use Source Prefix", which makes device acronyms unique, but was using UniqueID from existing device acronym.


ritchiecarroll:
> I guess this still doesn't fix the case where you update the device names manually.

No, guess not. However, in this case the record is highlighted as a duplicate / already existing, so there's user feedback on record being duplicated.

I'll see if I can handle this use case as well.

ritchiecarroll:
> I guess this still doesn't fix the case where you update the device names manually.

This should do it: https://github.com/GridProtectionAlliance/gsf/pull/261/commits/32c1b9a90c4a5bf9fe0ca5b1031d4cae8bf7d803

Any "new device", by manual entry or otherwise, gets unqiue ID vallidation.

Source/Libraries/Adapters/PhasorProtocolAdapters/CommonPhasorServices.cs
```diff
@@ -914,6 +915,8 @@ private void m_frameParser_ConnectionAttempt(object sender, EventArgs e)
     private static TemplatedExpressionParser s_pointTagExpressionParser;
     private static Dictionary<string, DataRow> s_signalTypes;
 
+    private static readonly string[] s_commonVoltageLevels = ["44", "69", "115", "138", "161", "169", "230", "345", "500", "765", "1100"];
```
StephenCWills, line 918: I recommend reversing the order here to help avoid issues, like how `"SHELBY_169KV".IndexOf("69")` would indicate that SHELBY is 69 kV instead of 169 kV.

```suggestion
    private static readonly string[] s_commonVoltageLevels = ["1100", "765", "500", "345", "230", "169", "161", "138", "115", "69", "44"];
```

Source/Libraries/GSF.PhasorProtocols/UI/WPF/ViewModels/InputWizardDevices.cs
```diff
@@ -56,2021 +56,2063 @@
...
-        /// <param name="sender">Source of the event.</param>
-        /// <param name="e">Event arguments.</param>
-        private void Helper_ReceivedServiceResponse(object sender, EventArgs<ServiceResponse> e)
+        if (m_configurationFrame != null)
+            cc.ConfigurationFrame = m_configurationFrame;
+
+        cc.Closed += delegate
         {
-            List<object> attachments = e.Argument.Attachments;
+            if (cc.DialogResult == null || !cc.DialogResult.GetValueOrDefault())
```
StephenCWills, line 1626:
```suggestion
            if (cc.DialogResult is null || !cc.DialogResult.GetValueOrDefault())
```

Source/Libraries/GSF.PhasorProtocols/UI/WPF/ViewModels/InputWizardDevices.cs
```diff
@@ -56,2021 +56,2063 @@
...
     }
-}
+
+    #endregion
+
+    #region [ Static ]
+
+    // Fields
+    private static ManualResetEvent s_responseWaitHandle;
+    private static readonly string[] s_commonVoltageLevels = ["44", "69", "115", "138", "161", "169", "230", "345", "500", "765", "1100"];
```
StephenCWills, line 2115: Same issue, different spot.

```suggestion
    private static readonly string[] s_commonVoltageLevels = ["1100", "765", "500", "345", "230", "169", "161", "138", "115", "69", "44"];
```

StephenCWills (changes requested):

StephenCWills (approved):

PR #298, opened 2024-10-03
GSF.Core: Split implementation of Middle() vs Median()

StephenCWills:
The name of the `Median()` function is misleading, which led to some sneaky logic errors in the islanding detection algorithm. Splitting the implementation like this should help to resolve potential confusion. I also snuck in some optimizations for non-array lists and collections which are not lists.

### DIFF
--- a/Source/Libraries/Adapters/GrafanaAdapters/Functions/BuiltIn/Median.cs
+++ b/Source/Libraries/Adapters/GrafanaAdapters/Functions/BuiltIn/Median.cs
@@ -43,18 +43,18 @@ public abstract class Median<T> : GrafanaFunctionBase<T> where T : struct, IData
         public override async IAsyncEnumerable<MeasurementValue> ComputeAsync(Parameters parameters, [EnumeratorCancellation] CancellationToken cancellationToken)
         {
             // Median uses immediate in-memory array load
-            MeasurementValue[] values = (await GetDataSourceValues(parameters).OrderBy(dataValue => dataValue.Value).ToArrayAsync(cancellationToken).ConfigureAwait(false)).Median();
-            int length = values.Length;
+            List<MeasurementValue> values = await GetDataSourceValues(parameters)
+                .OrderBy(dataValue => dataValue.Value)
+                .ToListAsync(cancellationToken)
+                .ConfigureAwait(false);
 
-            if (length == 0)
+            IEnumerable<MeasurementValue> median = values.Middle();
+
+            if (!median.Any())
                 yield break;
 
-            // Median can return two values if there is an even number of values
-            MeasurementValue result = values.Last();
-
-            if (length > 1)
-                result.Value = values[0].Value + (values[1].Value - values[0].Value) / 2.0D;
-
+            MeasurementValue result = median.Last();
+            result.Value = median.Average(mv => mv.Value);
             yield return result;
         }
     }
@@ -80,18 +80,9 @@ public abstract class Median<T> : GrafanaFunctionBase<T> where T : struct, IData
             if (magnitudes.Count == 0)
                 yield break;
 
-            double[] magnitudeMedians = magnitudes.OrderBy(value => value).Median();
-            double[] angleMedians = angles.OrderBy(value => value).Median();
-
-            double magnitudeMedian = magnitudeMedians.Last();
-            double angleMedian = angleMedians.Last();
-
             // Median can return two values if there is an even number of values
-            if (magnitudeMedians.Length > 1)
-                magnitudeMedian = magnitudeMedians[0] + (magnitudeMedians[1] - magnitudeMedians[0]) / 2.0D;
-
-            if (angleMedians.Length > 1)
-                angleMedian = angleMedians[0] + (angleMedians[1] - angleMedians[0]) / 2.0D;
+            double magnitudeMedian = magnitudes.Median().Average();
+            double angleMedian = angles.Median().Average();
 
             // Return computed results
             if (lastValue.Time > 0.0D)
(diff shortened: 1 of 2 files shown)


Source/Libraries/Adapters/GrafanaAdapters/Functions/BuiltIn/Median.cs
```diff
@@ -43,18 +43,18 @@ public class ComputeMeasurementValue : Median<MeasurementValue>
...
+                .OrderBy(dataValue => dataValue.Value)
+                .ToListAsync(cancellationToken)
+                .ConfigureAwait(false);
 
-            if (length == 0)
-                yield break;
-
-            // Median can return two values if there is an even number of values
-            MeasurementValue result = values.Last();
+            IEnumerable<MeasurementValue> median = values.Middle();
```
ritchiecarroll, line 51: Lot's of multiple enumerations even though it's two items, might as well just enumerate once:
```c#
MeasurementValue[] median = values.Middle().ToArray();
```

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
```diff
@@ -422,27 +422,69 @@ public static bool AllParallel<TSource>(this IEnumerable<TSource> source, Func<T
         /// <typeparam name="T"><see cref="Type"/> of elements in the <paramref name="source"/></typeparam>
         /// <param name="source">An enumeration over which to find the median element(s).</param>
         /// <returns>The median item(s) from an enumeration, or <c>null</c> if <paramref name="source"/> is <c>null</c>.</returns>
-        public static T[] Median<T>(this IEnumerable<T> source)
+        public static IEnumerable<T> Median<T>(this IEnumerable<T> source)
         {
             if (source == null)
```
ritchiecarroll, line 427: Better as `is null`, although the result of the functions in the files still use `== null`:
```c#
if (source is null)
```

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
```diff
@@ -422,27 +422,69 @@ public static bool AllParallel<TSource>(this IEnumerable<TSource> source, Func<T
...
+        /// <remarks>
+        /// If enumeration length is even, two items will be returned representing the items in the middle of the set;
+        /// otherwise, if the enumeration length is odd, one item will be returned from exactly the middle of the set.
+        /// </remarks>
+        /// <typeparam name="T"><see cref="Type"/> of elements in the <paramref name="source"/></typeparam>
+        /// <param name="source">An enumeration over which to find the middle element(s).</param>
+        /// <returns>The middle item(s) from an enumeration, or <c>null</c> if <paramref name="source"/> is <c>null</c>.</returns>
+        public static IEnumerable<T> Middle<T>(this IEnumerable<T> source)
+        {
+            if (source == null)
```
ritchiecarroll, line 447: Ditto

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
```diff
@@ -422,27 +422,69 @@ public static bool AllParallel<TSource>(this IEnumerable<TSource> source, Func<T
...
 
-            List<T> results = new List<T>();
+            // Instead of source.ToList(), we could use source.Count()
+            // followed by source.Skip(...).Take(...) to avoid memory overhead
+            // but at the cost of enumerating twice
+            if (source is IList<T> list)
+                return MiddleList(list);
+            if (source is ICollection<T> collection)
+                return MiddleCollection(collection);
+            else
```
ritchiecarroll, line 457: `else` is redundant here

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
```diff
@@ -422,27 +422,69 @@ public static bool AllParallel<TSource>(this IEnumerable<TSource> source, Func<T
...
+                return MiddleList(source.ToList());
+
+            IEnumerable<T> MiddleList(IList<T> list)
```
ritchiecarroll, line 460: Not sure on our official sub-function naming conventions - but I tend to make these camel cased (lately anyway)

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
```diff
@@ -422,27 +422,69 @@ public static bool AllParallel<TSource>(this IEnumerable<TSource> source, Func<T
...
+                    yield return list[midIndex - 1];
 
-            bool isEven = values.Length % 2 == 0;
-            int midIndex = values.Length / 2;
+                yield return list[midIndex];
+            }
 
-            if (isEven)
-                results.Add(values[midIndex - 1]);
+            IEnumerable<T> MiddleCollection(ICollection<T> collection)
```
ritchiecarroll, line 473: Camel?

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
ritchiecarroll: Also `list` hides already declared variable by same name above

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
ritchiecarroll: Ditto on variable hiding

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
ritchiecarroll: I suppose the distinction between `IList` and `ICollection` , since `ICollection` has a `Count` property and `IList` implements `ICollection`, is the lack of indexer in `ICollection`.

bah
```

ritchiecarroll (approved):
All these are nits - optional fixes - looks good.

Source/Libraries/GSF.Core/Collections/CollectionExtensions.cs
StephenCWills: I've been using PascalCase for local functions and continuing to use camelCase for delegate variables.

I don't know if there's any useful distinction between a local function and a non-local one when reading the callsite, as opposed to variables where you may need to know if you're modifying class state. By the same token, I'm not sure if there's any useful distinction between local function names and local delegate variables. 🤷

clackner-gpa (approved):

PR #301, opened 2024-10-04
DynamicCalculator: Add option for PublishWithBadState to support alarm inputs

StephenCWills:
Alarms only produce measurements when the alarm state changes so the published value will quickly expire become unusable for computations. In that case, we want to be able to use the latest value of the alarm regardless of how old it is.

I also removed superfluous logic in `ProcessMeasurements()`. This logic could never actually update `m_latestTimestamp` because it was already updated in `PublishFrame()`. Perhaps it was a leftover from an iteration that attempted to use `QueueMeasurementsForProcessing()` instead of `PublishFrame()`.


ritchiecarroll (commented):
Looks good

ritchiecarroll (dismissed):

StephenCWills:
Oh, perhaps I should add it to the adapter's status.

PR #313, opened 2024-11-18
GSF.InstallerActions: Add custom action for forcing read-only permissions on a folder

StephenCWills:


Source/Libraries/GSF.InstallerActions/CustomActions.cs
```diff
@@ -543,6 +544,26 @@ void addServiceAccountToGroup(string groupName)
...
+        /// <returns>Result of the custom action.</returns>
+        [CustomAction]
+        public static ActionResult ReadOnlyFolderAction(Session session)
+        {
+            Logger logger = new(session);
+
+            logger.Log("Begin ReadOnlyFolderAction");
+
+            string folderPath = GetPropertyValue(session, "FOLDERPATH");
+            SetFolderReadOnly(folderPath);
```
ritchiecarroll, line 560: What happens if `SetFolderReadOnly` throws an exception? Is it fatal to installation?

Should this be logged instead?
```c#
    Logger logger = new(session, "Windows Installer");
    logger.Log(EventLogEntryType.Warning, ex);
```

Not sure I know what the right answer should be - this is more a rhetorical question to make sure we are handling exception in the "correct" way for a custom action.

ritchiecarroll (approved):
I approve, but see comment.

Source/Libraries/GSF.InstallerActions/CustomActions.cs
StephenCWills: It should indeed be fatal, and the exception should be written to the log when using `/l*v`. I'm conflicted about handling the exception and making it nonfatal. On one hand, it would be annoying if there was an exception you couldn't resolve preventing you from completing the installation. On the other hand, the alternative is allowing a security vulnerability to slip through.

I'm not really sure what kind of error we could expect to encounter with this custom action, but we might be able to assume it should never fail. Alternatively, if we display a popup with `logger.Log(InstallMessage.Error, ...)`, it might be okay to let the installer continue.

PR #379, opened 2025-02-25
Added CSVAdapter for Collecting Frames

clackner-gpa:


StephenCWills:
Here's an overview for the changes I made.

* Update csproj file to include the new adapter
* Use `FramesPerSecond`, `LagTime`, and `LeadTime` parameters from base class instead of defining new ones
* Use rollover index to determine when to roll over
* Generate active file name based on rollover index instead of capturing it in a member variable
* Unless manually invoked, only offload files if `PublishFrame()` has had a chance to update rollover index
* Completely skip offload if the offload directory and export directory are the same

StephenCWills:
Since this has been in use at APG for a few days, I'm going to go ahead and merge it.

PR #387, opened 2025-03-20
Updated HTML bindings to escape HTML tags before rendering

ritchiecarroll:
Fixes script injection from read-only tag rendering.


StephenCWills (dismissed):
I guess we're moving away from ES5 completely? Not much sense in supporting IE these days, I suppose.

PR #398, opened 2025-04-11
DynamicCalculator: Adapter command to assist troubleshooting

StephenCWills:
This PR implements adapter commands to view the actual values of measurements received by the adapter as well as the values of variables used in the calculation of a given result. This is intended to assist with troubleshooting adapter configuration when the result of the calculation doesn't match expectations.

---

Here is a recommended approach for enabling these messages in a remote console application.
```
q
filter -include message MYCALC
r
invoke MYCALC RaiseVerboseMessages
```

Once the console has captured some of the output, you can disable the messages again with these commands.
```
q
invoke MYCALC StopVerboseMessages
```

---

Here is some example output from an adapter using the following calculation:
`If(freq >= 45 and freq <= 55 and volt >= 10000, 1, 0)`

FYI, the `x` variable is only included for timing and is not used in the calculation. In this case, the input device is providing frequency/voltage data at half the expected framerate, so the second result includes NaN values as input.

```
[MYCALC] Calculation details 2025-04-11 16:32:32.8400000:
freq = 49.9941139221191
LOCALTIME = 4/11/2025 6:32:34 PM
SYSTEMNAME = MYSERVER
TIME = 638799859528400000
UTCTIME = 4/11/2025 4:32:34 PM
volt = 408828.156482813
x = 0.244234193695818
Result = 1

[MYCALC] Calculation details 2025-04-11 16:32:32.8600000:
freq = NaN
LOCALTIME = 4/11/2025 6:32:34 PM
SYSTEMNAME = MYSERVER
TIME = 638799859528600000
UTCTIME = 4/11/2025 4:32:34 PM
volt = NaN
x = 0.230035456935892
Result = 0
```

### DIFF
--- a/Source/Libraries/Adapters/DynamicCalculator/DynamicCalculator.cs
+++ b/Source/Libraries/Adapters/DynamicCalculator/DynamicCalculator.cs
@@ -32,6 +32,7 @@ using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
+using System.Threading;
 using Ciloci.Flee;
 using GSF;
 using GSF.Collections;
@@ -117,6 +118,8 @@ namespace DynamicCalculator
         private IDynamicExpression m_expression;
 
         private readonly DelayedSynchronizedOperation m_timerOperation;
+
+        private int m_raisingVerboseMessages;
 
         #endregion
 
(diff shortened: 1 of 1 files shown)


Source/Libraries/Adapters/DynamicCalculator/DynamicCalculator.cs
```diff
@@ -629,7 +653,9 @@ protected void Calculate(IReadOnlyDictionary<MeasurementKey, IMeasurement> measu
             m_expression ??= m_expressionContext.CompileDynamic(m_aliasedExpressionText);
 
             // Evaluate the expression and generate the measurement
-            HandleCalculatedValue(m_expression.Evaluate());
+            object calculatedValue = m_expression.Evaluate();
+            HandleCalculatedValue(calculatedValue);
+            RaiseVerboseMessage(m_expressionContext.Variables, calculatedValue);
```
ritchiecarroll, line 658: This is just a comment - and the difference it makes would likely be microscopic - but instead of pushing parameters to the stack, then making the functional call, then in the function checking if you should raise verbose messages, would it not be optimal to check the flag first, skipping the stack prep and function call (both here and above)?

PR #404, opened 2025-05-18
Added extra Parameter to check Target of Reference

clackner-gpa:
Added an optional Parameter to specify Target of Reference and return NaN if it is not present


StephenCWills (approved):
It worked perfectly.

PR #410, opened 2025-05-28
Update GSF.Core.Console.Command.cs

ajstadlin:
Added SafeDisposal and CustomProcessStartInfo to GSF.Core.Console.Command.cs

- `SafeDispoal = true` tells the Dispose method to Close and Kill the process before disposing it.
    - Default is `SafeDispoal = false` so existing implementations are not affected by this new property.

- `CustomProcessStartInfo` enables setting or modifying additional properties in the Process.StartInfo.
    - Default is `CustomProcessStartInfo = null` so existing implementations are not affected by this new property.

- No original code removed

//AJS


ritchiecarroll (approved):
Hi AJ! Looks great!

PR #414, opened 2025-06-20
CSVDataManager: Improve exception handling

StephenCWills:
Here are some screenshots from a couple failed imports using the changes in this PR.

---

image

---

image

image


Source/Tools/CSVDataManager/MainWindow.cs
```diff
@@ -727,6 +761,38 @@ private void UpdateProgressBar(ProgressBar progressBar, int progress)
                 updateAction(progressBar, progress);
         }
 
+        private void ShowErrors()
+        {
+            if (!MainTabControl.TabPages.Contains(ErrorsTabPage))
+                MainTabControl.TabPages.Add(ErrorsTabPage);
```
ritchiecarroll, line 767: I presume adding errors tab shows it / makes it active? If function should always "show", should it not also make this selected tab in case tab already exists? Then, in that case line 439 could just call `ShowErrors`.

If the function purpose is to otherwise ensure the tab exists perhaps a better name would be something like `AddErrorsTabPage`

ritchiecarroll (changes requested):

Source/Tools/CSVDataManager/MainWindow.cs
StephenCWills: > If the function purpose is to otherwise ensure the tab exists perhaps a better name would be something like `AddErrorsTabPage`

Calling `Control.Hide()` and `Control.Show()` do nothing when invoked on a tab page, so I am removing it and adding it back instead which is effectively the same as what I intended with hide/show. I named the functions `HideErrors()` and `ShowErrors()` because it aligns with my original intent, even though I had to do it in a different way. Honestly, this is technically unnecessary, but I think it does help a bit with UX to only display the tab when relevant.

> I presume adding errors tab shows it / makes it active? If function should always "show", should it not also make this selected tab in case tab already exists? Then, in that case line 439 could just call `ShowErrors`.

Fair point. There's no real reason to assume we won't be switching to the tab page when `ShowErrors()` is called.

PR #517, opened 2026-04-23
Improve immediate measurements

ritchiecarroll:
Adds `HasMeasurementID` function to see if immediate measurement cache has desired `MeasurementKey`


StephenCWills (approved):

StephenCWills:
FYI, I convinced Ritchie to close the PR, thinking that `immediateMeasurements.MeasurementIDs.Contains()` would be equivalent to `HasMeasurementIDs()`, but it turns out that's only true for `Dictionary<TKey, TValue>`. The `ConcurrentDictionary<TKey, TValue>` implementation actually creates a copy of the keys as an array and wraps it in a `ReadOnlyCollection<TKey>`. So in this case, it's much more efficient to use `ContainsKey()` directly.

PR #521, opened 2026-04-23
GSF-17 Correct parsing logic for common expressions in TemplatedExpressionParser

StephenCWills:
This PR fixes an issue where a template with a common expression that has square brackets nested within it would not parse correctly.

I tested using the program below. Before the changes in this PR, I received the following error.

> ExpressionEvaluator.Parser.ExpressionParseException: 'Error parsing token '7' at line 1 char 26'

The expected result is `DAMP.MODE50.!.MAG`.

```c#
using System;
using System.Collections.Generic;
using GSF.Parsing;

namespace TemplatedExpressionParserTest
{
    internal class Program
    {
        static void Main(string[] args)
        {
            TemplatedExpressionParser parser = new TemplatedExpressionParser();
            parser.TemplatedExpression = @"[?{Output}=ModeShapeMagnitude[DAMP.MODE{ModeNum}.eval{'{Input.PointTag}'[7]}.MAG]]";

            Dictionary<string, string> substitutions = new Dictionary<string, string>()
            {
                { "{Output}", "ModeShapeMagnitude" },
                { "{ModeNum}", "50" },
                { "{Input.PointTag}", "GPA_PDC!SHELBY:FREQ" }
            };

            string result = parser.Execute(substitutions);
            Console.WriteLine(result);
        }
    }
}
```

### DIFF
--- a/Source/Libraries/GSF.Core/Parsing/TemplatedExpressionParser.cs
+++ b/Source/Libraries/GSF.Core/Parsing/TemplatedExpressionParser.cs
@@ -35,7 +35,6 @@ using System.Text.RegularExpressions;
 #if DNF46
 using ExpressionEvaluator;
 #endif
-using ParsedExpression = System.Tuple<string, bool, string>;
 using ParsedEvaluation = System.Tuple<string, string>;
 
 namespace GSF.Parsing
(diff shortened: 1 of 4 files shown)


Source/Libraries/GSF.Core/Parsing/TemplatedExpressionParser.cs
ritchiecarroll, line 1: Took me a bit to parse this -- main changes I think are these: cumulative "AND" is now recursion-based (the outer must succeed before the inner is evaluated), eliminating the evaluations/depth tracking. Uses new balanced regex. `ParseExpressions` simplified by returning a string instead of applying replace pass on every list item.

This looks OK, but was a big change, so I added unit tests to this PR that cover common and more complex example template cases, plus the new bracket case the update was intending to handle -- these all pass -- so I feel very good about this update.

There is one small breaking _behavior_ change:  I think the old code would silently produce an empty list and drop expression text on malformed templates -- the new code throws `InvalidOperationException` with a "check balance of square brackets" message.  I think the new implementation is arguably more correct, but it might be a breaking change for any consumer that otherwise might have been tolerating a bad template -- something to keep in mind / watch for after update -- but I don't think it's a huge concern, and I could name off the users with templates I might be concerned with...

On small thing you could update:  the `[SuppressMessage("Microsoft.Maintainability", "CA1502")]` on the new `ParseExpressions` can be dropped, the method is short now...

ritchiecarroll (changes requested):
Minor change requested.

ritchiecarroll (approved):

PR #524, opened 2026-04-28
Add optional power factor calculation to PowerMultiCalculator

ritchiecarroll:
Adds two new connection-string parameters to `PowerMultiCalculatorAdapter`:
`EnablePowerFactorCalculation` (default `false`, preserves existing behavior) and `PowerFactorSignConvention` (`Unsigned`/`Lagging`/`Leading`, default `Unsigned`).

When enabled, per-circuit "-PF" output measurements are auto-created via GSF.TimeSeries.Adapters.MetadataHelpers using the existing MW/MVA/MVAR point tag as a naming anchor, with no changes required to the PowerCalculation database table, validator, or UI. PF is computed inline in PublishFrame as |P|/|S|, optionally signed by Q to encode lagging or leading direction. Per-circuit failures soft-fail with a warning so misconfigured circuits do not prevent the rest of the adapter from running.

### DIFF
--- a/Source/Libraries/Adapters/PowerCalculations/PowerMultiCalculator/PowerMultiCalculatorAdapter.cs
+++ b/Source/Libraries/Adapters/PowerCalculations/PowerMultiCalculator/PowerMultiCalculatorAdapter.cs
@@ -41,6 +41,7 @@ using GSF.TimeSeries;
 using GSF.TimeSeries.Adapters;
 using GSF.Units;
 using GSF.Units.EE;
+using MeasurementRecord = GSF.TimeSeries.Model.Measurement;
 
 namespace PowerCalculations.PowerMultiCalculator;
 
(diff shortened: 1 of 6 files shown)


Source/Libraries/Adapters/PowerCalculations/PowerMultiCalculator/PowerMultiCalculatorAdapter.cs
```diff
@@ -547,6 +673,197 @@ bool includeInput(IMeasurement measurement)
...
+        if (recordsAdded)
+        {
+            // Notify host system that configuration has changed
+            this.OnConfigurationChanged();
+
+            OnStatusMessage(MessageLevel.Info, "Waiting for the newly created power factor output measurements to be loaded into active configuration...");
+
+            if (!this.WaitForSignalsToLoad(m_configurationReloaded, newSignalIDs))
+                OnStatusMessage(MessageLevel.Warning, $"Power factor output measurements not found in active configuration after waiting {MetadataHelpers.ElapsedWaitTimeString()} - power factor outputs may not be available until next reload.");
+        }
```
StephenCWills, line 786: Seems like we should reset `m_configurationReloaded` before `this.OnConfigurationChanged()` so that `this.WaitForSignalsToLoad()` is guaranteed to receive the new signals.
```suggestion
        if (recordsAdded)
        {
            m_configurationReloaded.Reset();

            // Notify host system that configuration has changed
            this.OnConfigurationChanged();

            OnStatusMessage(MessageLevel.Info, "Waiting for the newly created power factor output measurements to be loaded into active configuration...");

            if (!this.WaitForSignalsToLoad(m_configurationReloaded, newSignalIDs))
                OnStatusMessage(MessageLevel.Warning, $"Power factor output measurements not found in active configuration after waiting {MetadataHelpers.ElapsedWaitTimeString()} - power factor outputs may not be available until next reload.");
        }
```

ritchiecarroll: Seems safe enough

StephenCWills (approved):

PR #527, opened 2026-05-04
Fix logic when rounding to second distribution

StephenCWills:
This PR corrects a couple issues with the previous implementation of `Ticks.RoundToSecondDistribution()`.

* Passing a baseline timestamp of 0 would inexplicably cause a divide-by-zero error despite the fact that it should be a valid (and common) starting timestamp.
* Passing a framerate with a reciprocal that is not divisible by the baseline would cause "hiccups" in the distribution, where the end of every "baseline interval" would be shorter than it's supposed to be.

To simplify the logic, I tried to conceptualize the operation of converting between timestamp and sample number as a linear equation that passes through the baseline at frame index 0. The comments in the code describe the mathematical model and how it is applied to produce the rounded result.

The further your timestamp is from the baseline, the more likely you are to lose precision in the computation because that time difference (`timestamp - baseline`) is converted to a floating-point number--from 64 bits of precision down to 53. Notably, `DateTime.UtcNow` already returns a number with 60 bits of precision, which would cause 7 bits of precision to be lost when using a baseline of 0. This can be mitigated by picking a different baseline closer to the given timestamp that produces the same result, which can be done by adding an integer multiple of the sampling interval to the given baseline. This is a bit tricky, since the sample rate--and by extension, its reciprocal--can be any floating-point number, but it's done quickly and precisely by simply extracting the mantissa to an integer, as if we had multiplied the number by 2 until the exponent is equal to the maximum precision of the floating-point representation.

---

I tested this code using the following test app.

```c#
using System;
using GSF;

namespace SecondDistributionTest
{
    internal class Program
    {
        static void Main(string[] args)
        {
            Ticks now = DateTime.UtcNow.Ticks;
            Ticks round0 = Ticks.RoundToSecondDistribution(now, 1.0D / 1.5D, 0L * Ticks.PerSecond);
            Ticks round1 = Ticks.RoundToSecondDistribution(now, 1.0D / 1.5D, 1L * Ticks.PerSecond);
            Ticks round2 = Ticks.RoundToSecondDistribution(now, 1.0D / 1.5D, 2L * Ticks.PerSecond);
            Ticks round3 = Ticks.RoundToSecondDistribution(now, 1.0D / 1.5D, 3L * Ticks.PerSecond);

            // These should all be within 1.5 seconds
            double diff0 = TimeSpan.FromTicks(now - round0).TotalSeconds;
            double diff1 = TimeSpan.FromTicks(now - round1).TotalSeconds;
            double diff2 = TimeSpan.FromTicks(now - round2).TotalSeconds;
            double diff3 = TimeSpan.FromTicks(now - round3).TotalSeconds;

            // These should all be zero
            long interval = TimeSpan.FromSeconds(1.5D).Ticks;
            long mod0 = (round0 - 0L * Ticks.PerSecond) % interval;
            long mod1 = (round1 - 1L * Ticks.PerSecond) % interval;
            long mod2 = (round2 - 2L * Ticks.PerSecond) % interval;
            long mod3 = (round3 - 3L * Ticks.PerSecond) % interval;

            Console.WriteLine(ToString(now));
            Console.WriteLine($"{ToString(round0)} {diff0} {mod0}");
            Console.WriteLine($"{ToString(round1)} {diff1} {mod1}");
            Console.WriteLine($"{ToString(round2)} {diff2} {mod2}");
            Console.WriteLine($"{ToString(round3)} {diff3} {mod3}");
        }

        static string ToString(Ticks timestamp) =>
            $"{timestamp:yyyy-MM-dd HH:mm:ss.fffffff}";
    }
}
```

### DIFF
--- a/Source/Libraries/GSF.Core.Shared/Ticks.cs
+++ b/Source/Libraries/GSF.Core.Shared/Ticks.cs
@@ -1629,33 +1629,66 @@ namespace GSF
         /// </summary>
         /// <param name="timestamp">Timestamp to align.</param>
         /// <param name="samplesPerSecond">Samples per second to use for distribution.</param>
-        /// <param name="Baseline"> Starting Timestamp of the Distribution.</param>
+        /// <param name="baseline">Starting timestamp of the distribution.</param>
         /// <returns>The nearest distribution timestamp for given <paramref name="timestamp"/>.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static Ticks RoundToSecondDistribution(Ticks timestamp, double samplesPerSecond, Ticks Baseline)
+        public static Ticks RoundToSecondDistribution(Ticks timestamp, double samplesPerSecond, Ticks baseline)
         {
-            // Calculate destination ticks for this frame
-            long ticks = timestamp.Value;
-            long baseTicks, ticksBeyondSecond, destinationTicks;
-            double frameIndex;
+            if (double.IsNaN(samplesPerSecond) || double.IsInfinity(samplesPerSecond))
+                throw new ArgumentOutOfRangeException(nameof(samplesPerSecond));
 
-            // Baseline timestamp based on start Time
-            baseTicks = ticks - ticks % Baseline;
+            if (samplesPerSecond <= 0.0D)
+                throw new ArgumentOutOfRangeException(nameof(samplesPerSecond));
 
-            // Remove the seconds from ticks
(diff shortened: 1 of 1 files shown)


StephenCWills:
FYI, I did a little math to figure out if `ticksPerSample < (1L << 53)` is too restrictive. `1*2^53 ticks` is a little over 28.5 years. So as long as the sampling interval is 25 years or less, this restriction should be okay.

ritchiecarroll (approved):

PR #528, opened 2026-05-04
Dnp3Adapters: Update Dnp3InputAdapter to accept polling intervals by class

StephenCWills:
The new connection string syntax is a semicolon-separated list of key-value pairs. The key is a comma-separated list of classes to be polled at a given polling interval and the value is a polling interval specified in seconds. Each of these key-value pairs will be turned into a single invocation of `master.AddClassScan()`.

If the new `PollingIntervals` parameter is not specified, the adapter will continue to look for the old `PollingInterval` parameter and convert the value into an expression of the form `Class0,Class1,Class2,Class3={PollingInterval}`. This should match the original behavior of the adapter so old connection strings don't need to be modified unless the user wishes to change the behavior.

Likewise, the default value is `Class0,Class1,Class2,Class3=2` to match old behavior as well, although there is a case to be made for removing `Class0` from the default.

### DIFF
--- a/Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
+++ b/Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
@@ -152,11 +152,25 @@ public class DNP3InputAdapter : InputAdapterBase, IDnp3Adapter
     }
 
     /// <summary>
-    /// Gets or sets the string representation of the list of intervals
-    /// at which the adapter will poll each class of data from the DNP3 device.
+    /// Gets or sets the string representation of the list of intervals, in seconds, at which the
+    /// adapter will poll each class of data from the DNP3 device.
     /// </summary>
+    /// <remarks>
+    /// The string format for the polling intervals is a semicolon-separated list of intervals where
+    /// each interval is a comma separated list of class names followed by an equals sign and the
+    /// interval in seconds. For example: <c>Class0,Class1=2; Class2=10; Class3=30</c> would configure
+    /// the adapter to poll Class 0 and Class 1 every 2 seconds, Class 2 every 10 seconds, and Class 3
+    /// every 30 seconds. The class names should be one of 'Class0', 'Class1', 'Class2', or 'Class3'.
+    /// </remarks>
     [ConnectionStringParameter]
-    [Description("Define the semicolon-separated list of intervals, in seconds, at which the adapter will poll each class of data from the DNP3 device.")]
+    [Description(
+        "Define the semicolon-separated list of intervals, in seconds, at which the adapter will " +
+        "poll each class of data from the DNP3 device. The string format for the polling intervals " +
+        "is a semicolon-separated list of intervals where each interval is a comma separated list " +
+        "of class names followed by an equals sign and the interval in seconds.\r\n\r\nFor example: " +
+        "\"Class0,Class1=2; Class2=10; Class3=30\" would configure the adapter to poll Class 0 and " +
+        "Class 1 every 2 seconds, Class 2 every 10 seconds, and Class 3 every 30 seconds. The class " +
+        "names should be one of 'Class0', 'Class1', 'Class2', or 'Class3'.")]
     [DefaultValue(DefaultPollingIntervals)]
     public string PollingIntervals
     {
@@ -331,8 +345,8 @@ public class DNP3InputAdapter : InputAdapterBase, IDnp3Adapter
 
         IMaster master = channel.AddMaster(ChannelID, m_soeHandler, DefaultMasterApplication.Instance, m_masterConfig.master);
 
-        foreach ((ClassField, TimeSpan) pollingInterval in m_pollingIntervalList)
-            master.AddClassScan(pollingInterval.Item1, pollingInterval.Item2, m_soeHandler, TaskConfig.Default);
+        foreach ((ClassField classField, TimeSpan interval) pollingInterval in m_pollingIntervalList)
+            master.AddClassScan(pollingInterval.classField, pollingInterval.interval, m_soeHandler, TaskConfig.Default);
 
         master.Enable();
         m_active = true;
@@ -416,7 +430,7 @@ public class DNP3InputAdapter : InputAdapterBase, IDnp3Adapter
             .Select(ParsePollingInterval)
             .ToList();
 
-        (ClassField ClassField, TimeSpan Interval) ParsePollingInterval(string pollingInterval)
+        (ClassField, TimeSpan) ParsePollingInterval(string pollingInterval)
         {
             string[] parts = pollingInterval.Split('=');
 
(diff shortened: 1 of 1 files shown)


Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
```diff
@@ -332,8 +331,8 @@ protected override void AttemptConnection()
 
         IMaster master = channel.AddMaster(ChannelID, m_soeHandler, DefaultMasterApplication.Instance, m_masterConfig.master);
 
-        if (m_pollingInterval > TimeSpan.Zero)
-            master.AddClassScan(ClassField.AllClasses, m_pollingInterval, m_soeHandler, TaskConfig.Default);
+        foreach ((ClassField, TimeSpan) pollingInterval in m_pollingIntervalList)
+            master.AddClassScan(pollingInterval.Item1, pollingInterval.Item2, m_soeHandler, TaskConfig.Default);
```
ritchiecarroll: This might be a good place to use named tuple elements for clarity:
```suggestion
        foreach ((ClassField classField, TimeSpan interval) pollingInterval in m_pollingIntervalList)
            master.AddClassScan(pollingInterval.classField, pollingInterval.interval, m_soeHandler, TaskConfig.Default);
```

Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
```diff
@@ -162,15 +152,20 @@ public double TimestampDifferentiation
...
     /// <summary>
-    /// Gets or sets the interval, in seconds, at which the adapter will poll the DNP3 device.
+    /// Gets or sets the string representation of the list of intervals
+    /// at which the adapter will poll each class of data from the DNP3 device.
     /// </summary>
     [ConnectionStringParameter]
-    [Description("Define the interval, in seconds, at which the adapter will poll the DNP3 device.")]
-    [DefaultValue(DefaultPollingInterval)]
-    public double PollingInterval
+    [Description("Define the semicolon-separated list of intervals, in seconds, at which the adapter will poll each class of data from the DNP3 device.")]
```
ritchiecarroll: This is a more complex connection string format, suggest improving comment / description for implementation clarity:
```suggestion
    /// <summary>
    /// Gets or sets the string representation of the list of intervals, in seconds, at which the
    /// adapter will poll each class of data from the DNP3 device.
    /// </summary>
    /// <remarks>
    /// The string format for the polling intervals is a semicolon-separated list of intervals where
    /// each interval is a comma separated list of class names followed by an equals sign and the
    /// interval in seconds. For example: <c>Class0,Class1=2; Class2=10; Class3=30</c> would configure
    /// the adapter to poll Class 0 and Class 1 every 2 seconds, Class 2 every 10 seconds, and Class 3
    /// every 30 seconds. The class names should be one of 'Class0', 'Class1', 'Class2', or 'Class3'.
    /// </remarks>
    [ConnectionStringParameter]
    [Description(
        "Define the semicolon-separated list of intervals, in seconds, at which the adapter will " +
        "poll each class of data from the DNP3 device. The string format for the polling intervals " +
        "is a semicolon-separated list of intervals where each interval is a comma separated list " +
        "of class names followed by an equals sign and the interval in seconds.\r\n\r\nFor example: " + 
        "\"Class0,Class1=2; Class2=10; Class3=30\" would configure the adapter to poll Class 0 and " +
        "Class 1 every 2 seconds, Class 2 every 10 seconds, and Class 3 every 30 seconds. The class " +
        "names should be one of 'Class0', 'Class1', 'Class2', or 'Class3'.")]
```

Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
```diff
@@ -408,5 +407,44 @@ private static T ReadConfig<T>(string path)
...
+    private static List<(ClassField, TimeSpan)> ParsePollingIntervals(string text)
+    {
+        return text
+            .Split(';')
+            .Select(interval => interval.Trim())
+            .Where(interval => !string.IsNullOrEmpty(interval))
+            .Select(ParsePollingInterval)
+            .ToList();
+
+        (ClassField ClassField, TimeSpan Interval) ParsePollingInterval(string pollingInterval)
```
ritchiecarroll: These named tuple elements are not used, can drop:
```suggestion
        (ClassField, TimeSpan) ParsePollingInterval(string pollingInterval)
```

Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
```diff
@@ -408,5 +407,44 @@ private static T ReadConfig<T>(string path)
...
+                .Select(ParsePointClass)
+                .ToArray();
+
+            ClassField classField = ClassField.From(pointClasses);
+
+            if (!double.TryParse(parts[1], out double seconds))
+                throw new FormatException($"Expected interval number of seconds, got {parts[1]}");
+
+            TimeSpan timeSpan = TimeSpan.FromSeconds(seconds);
+            return (classField, timeSpan);
```
ritchiecarroll: Suggest renaming `timeSpan` to `interval` for consistency / clarity:
```suggestion
            TimeSpan interval = TimeSpan.FromSeconds(seconds);
            return (classField, interval);
```

Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
```diff
@@ -408,5 +407,44 @@ private static T ReadConfig<T>(string path)
...
+        }
+
+        PointClass ParsePointClass(string text)
+        {
+            if (!Enum.TryParse(text, out PointClass pointClass))
+                throw new FormatException($"Expected point class, got {text}");
```
ritchiecarroll: The parameter `text` in this sub-function overrides the outer function parameter with the same name, to avoid confusion to the reader, suggest renaming:
```suggestion
        PointClass ParsePointClass(string className)
        {
            if (!Enum.TryParse(className, true, out PointClass pointClass))
                throw new FormatException($"Expected valid point class name, got {className}");
```

Suggestion also adds `true` parameter to `Enum.TryParse` to support case-insensitive class definitions, e.g.: `class0=60;Class1=2`, etc.

ritchiecarroll (changes requested):
Requesting a few changes.

Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
StephenCWills: The middle parameter is `ignoreCase`, so if you want case-insensitive matching, we should pass `true` as the second parameter.

Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
ritchiecarroll: Alternately, this should work too, and is more concise:
```suggestion
        foreach ((classField, interval) in m_pollingIntervalList)
            master.AddClassScan(classField, interval, m_soeHandler, TaskConfig.Default);
```

Source/Libraries/Adapters/Dnp3Adapters/Dnp3InputAdapter.cs
ritchiecarroll: Ah, right, that's what I was going for...

ritchiecarroll (approved):

PR #531, opened 2026-05-12
GSF.TimeSeries: Move filter adapter processing into routing tables

StephenCWills:
The strategy is:
1. Remove code from `IaonSession` that passes measurements into the filter adapters collection.
2. Expose the filter adapters collection to the routing tables.
3. Pass filter adapters into the routing tables' patch method.
4. Within the routing tables, add filter adapters into the global cache.
5. Do not invoke `IFilterAdapter.HandleNewMeasurements()` on the input adapter's `NewMeasurements` handler's thread!

In `RouteMappingDoubleBufferQueue`, this means we needed to introduce an `AsyncQueue<ICollection<IMeasurement>>` to pass the measurements into if there are filter adapters that need to process the measurements. Ideally, we'd be able to skip the async queue if there are no filter adapters, but this turned out to be a bit complicated because measurements from the input adapter could get published to consumers out of order if all filter adapters are disabled while the async queue has a backlog. Thus, I introduced a new field into the local cache for tracking whether there are any pending async jobs. That way, we can publish to the async queue even when there are no filter adapters, just to keep processing measurements in the correct order.

`RouteMappingHighLatencyLowCpu` was already doing all of its routing on a separate thread via the `ScheduledTask` class so the changes were much easier.

### DIFF
--- a/Source/Libraries/GSF.TimeSeries/Adapters/RoutingTables.cs
+++ b/Source/Libraries/GSF.TimeSeries/Adapters/RoutingTables.cs
@@ -116,6 +116,11 @@ namespace GSF.TimeSeries.Adapters
         public InputAdapterCollection InputAdapters { get; set; }
 
         /// <summary>
+        /// Gets or sets the active <see cref="FilterAdapterCollection"/>.
+        /// </summary>
+        public FilterAdapterCollection FilterAdapters { get; set; }
+
+        /// <summary>
         /// Gets or sets the active <see cref="ActionAdapterCollection"/>.
         /// </summary>
         public ActionAdapterCollection ActionAdapters { get; set; }
@@ -188,6 +193,7 @@ namespace GSF.TimeSeries.Adapters
             long startTime = DateTime.UtcNow.Ticks;
 
             IInputAdapter[] inputAdapterCollection = null;
+            IFilterAdapter[] filterAdapterCollection = null;
             IActionAdapter[] actionAdapterCollection = null;
             IOutputAdapter[] outputAdapterCollection = null;
             bool retry = true;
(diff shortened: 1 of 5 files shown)


Source/Libraries/GSF.TimeSeries/Adapters/RoutingTables.cs
```diff
@@ -238,10 +248,13 @@ private void CalculateRoutingTables()
                 HashSet<IAdapter> consumerAdapters = new((actionAdapterCollection ?? Enumerable.Empty<IAdapter>())
                     .Concat(outputAdapterCollection ?? Enumerable.Empty<IAdapter>()));
 
+                // Put filter adapters in execution order
+                Array.Sort(filterAdapterCollection, (adapter1, adapter2) => adapter1.ExecutionOrder.CompareTo(adapter2.ExecutionOrder));
```
ritchiecarroll:
```suggestion
                if (filterAdapterCollection is not null)
                    Array.Sort(filterAdapterCollection, (adapter1, adapter2) => adapter1.ExecutionOrder.CompareTo(adapter2.ExecutionOrder));
```

I think if there are no filter adapters defined this can be `null`

```
Application Domain:    openPDC.exe
Assembly Codebase:     C:/Program Files/openPDC/openPDC.exe
Assembly Full Name:    openPDC, Version=2.9.460.0, Culture=neutral, PublicKeyToken=null
Assembly Version:      2.9.460.0
Assembly Build Date:   4/30/2026 3:12:44 AM
.Net Runtime Version:  4.0.30319.42000

Exception Source:      mscorlib
Exception Type:        System.ArgumentNullException
Exception Message:     Value cannot be null.
Parameter name: array
Exception Target Site: Sort

---- Stack Trace ----
   System.Array.Sort(array As T[], comparison As Comparison`1)
       openPDC.exe: N 6584650
   GSF.TimeSeries.Adapters.RoutingTables.CalculateRoutingTables()
       RoutingTables.cs: Ln 0252, Col 17, IL 0331

```

ritchiecarroll (commented):
So far, found this, still testing...

ritchiecarroll (approved):

